PR #3, opened 2015-09-06
Fix returning argument types from methods.

ngrewe:
A tiny fix: method_copyArgumentType() didn't account for the fact that the return type is the first element in the method signature.

### DIFF
--- a/encoding2.c
+++ b/encoding2.c
@@ -433,7 +433,7 @@ unsigned method_get_number_of_arguments(struct objc_method *method)
 char* method_copyArgumentType(Method method, unsigned int index)
 {
 	if (NULL == method) { return NULL; }
-	const char *types = findParameterStart(method->types, index);
+	const char *types = findParameterStart(method->types, index + 1);
 	if (NULL == types)
 	{
 		return NULL;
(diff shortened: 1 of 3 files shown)


encoding2.c
```diff
@@ -433,7 +433,7 @@ unsigned method_get_number_of_arguments(struct objc_method *method)
 char* method_copyArgumentType(Method method, unsigned int index)
 {
 	if (NULL == method) { return NULL; }
-	const char *types = findParameterStart(method->types, index);
+	const char *types = findParameterStart(method->types, index + 1);
```
davidchisnall: Does `method_getArgumentType` need the same fix?  And please can we have a test case?

ngrewe:
> Does `method_getArgumentType` need the same fix?

It did indeed.

> And please can we have a test case?

Included now.

encoding2.c
```diff
@@ -393,7 +393,7 @@ void method_getArgumentType(Method method,
                             size_t dst_len)
 {
 	if (NULL == method) { return; }
-	const char *types = findParameterStart(method->types, index);
+	const char *types = findParameterStart(method->types, index + 1);
```
davidchisnall: Rather than having the +1 in both places, it would probably be cleaner to fix `findParaneterStart` to start in the right place.

ngrewe:
Good idea, thanks. Fixed now.

PR #4, opened 2015-09-07
Fix throwing nil exceptions.

ngrewe:
The following is perfectly legal on Mac OS X:

```objc
@try
{
  @throw(nil);
}
@catch (id obj)
{
  printf("Caught: %p", obj);
}
```
It doesn't work presently because we were trying to dereference the object pointer to determine the class.

### DIFF
--- a/Test/NilException.m
+++ b/Test/NilException.m
@@ -1,5 +1,23 @@
 #include "Test.h"
 
+
+#ifdef __has_attribute
+#if __has_attribute(objc_root_class)
+__attribute__((objc_root_class))
+#endif
+#endif
+@interface NSObject
+{
+  Class isa;
+}
+@end
+
+@implementation NSObject
++ (id)new
+{
+	return class_createInstance(self, 0);
+}
+@end
 int main(void)
 {
 	BOOL caught_exception = NO;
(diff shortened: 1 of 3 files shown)


Test/NilException.m
```diff
@@ -0,0 +1,17 @@
...
+	BOOL caught_exception = NO;
+	@try
+	{
+		@throw(nil);
+	}
+	@catch (id x)
+	{
+		assert(nil == x);
+		caught_exception = YES;
+	}
```
davidchisnall: What's the correct behaviour for, throwing `nil` and catching  (for example)`NSObject *`?  Should `nil` also be caught by `void*` catches in Objective-C++ (this case used to crash the Apple runtime, not sure if they've fixed it now)?

ngrewe:
> What's the correct behaviour for, throwing nil and catching (for example)NSObject *? 

It is only caught by @catch(id) and @catch(...), not by any other interface type. I've added a few more tests for those cases, but I'm running into some weird behaviour here: The standalone test case passes, but similar code, linked against gnustep-base doesn't work (i.e. @catch(NSObject*) catches nil as well…

> Should nil also be caught by void* catches in Objective-C++ (this case used to crash the Apple runtime, not sure if they've fixed it now)?

I've checked and it's not caught by a void* handler on 10.10.5 (it doesn't crash either, fwiw).

davidchisnall:
The stand-alone code may be interacting oddly with the CXXException.mm stuff in -base, which attempts to box C++ exceptions as Objective-C objects.  We should probably remove that now - it was largely a work-around for the fact that we didn't have working Objective-C++.

PR #15, opened 2016-02-08
Fix a segfault on (s|g)etting ivars.

ngrewe:
When we were trying to get or set an ivar on a class that had
no ivars defined, we'd dereference a NULL ivar list while
determining the ownership qualifier.

I've fixed that, and extended the test case to cover this case (and the case where the receiver has a non-NULL ivar list)

### DIFF
--- a/Test/ivar_arc.m
+++ b/Test/ivar_arc.m
@@ -20,14 +20,29 @@ int dealloc = 0;
 }
 @end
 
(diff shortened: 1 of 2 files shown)


davidchisnall:
Thanks.  I think there are a few cases in `runtime.c` where we don't do adequate `NULL` checking, which could possibly benefit from some test cases and/or visual inspection.

Issue #20, opened 2016-10-20
Deadlock in objc_send_initialize()


ngrewe:
Hi David,

We've discovered a fun little deadlock issue related to class initialisation. Assume two threads (1,2) and two classes (A,B) where A and B are unrelated, and A's +initialize sends a message to B:

1. Thread 1 runs and starts invoking objc_send_initialize([A class])
2. Thread 2 runs and starts invoking objc_send_initialize([A class])
3. Thread 1 is preempted by a high-priority thread when it reaches [dtable.c:855](https://github.com/gnustep/libobjc2/blob/9a2f43ca7d579928279b5d854a19adfbe43d06d6/dtable.c#L855) (right before the objc_class_flag_initialized flag is set). At this stage, it holds both the class and the runtime lock
4. Thread 2 continues running, past the block [dtable.c:830–839](https://github.com/gnustep/libobjc2/blob/9a2f43ca7d579928279b5d854a19adfbe43d06d6/dtable.c#L830-L839) (which tests the objc_class_flag_initialized flag for an early return), waiting for the runtime lock at [dtable.c:846](https://github.com/gnustep/libobjc2/blob/9a2f43ca7d579928279b5d854a19adfbe43d06d6/dtable.c#L846)
5. Thread 1 is rescheduled and relinquishes the runtime lock at [dtable.c:871](https://github.com/gnustep/libobjc2/blob/9a2f43ca7d579928279b5d854a19adfbe43d06d6/dtable.c#L871) it still holds the class lock
6. Thread 2 acquires the runtime lock, and blocks on the class lock at [dtable.c:847](https://github.com/gnustep/libobjc2/blob/9a2f43ca7d579928279b5d854a19adfbe43d06d6/dtable.c#L847)
7. Thread 1 continues running to invoke A's +initialize method and subsequently objc_send_initialize([B class]) and starts waiting on the runtime lock at [dtable.c:846](https://github.com/gnustep/libobjc2/blob/9a2f43ca7d579928279b5d854a19adfbe43d06d6/dtable.c#L846)

Now you have a deadlock: Thread 1 is waiting for thread 2 to release the runtime lock and thread 2 is waiting for thread 1 to release the class lock.

I think it might be a solution to acquire the runtime lock before the objc_class_flag_initialized test on line 830 (and unlocking it right after taking the branch, before acquiring the class lock).

davidchisnall:
[...]

PR #21, opened 2016-10-24
Avoid deadlock in objc_send_initialize()

ngrewe:
See gnustep/libobjc2#20

### DIFF
--- a/dtable.c
+++ b/dtable.c
@@ -824,6 +824,13 @@ PRIVATE void objc_send_initialize(id object)
 		objc_send_initialize((id)class->super_class);
 	}
 
+	// Lock the runtime while we're creating dtables and before we acquire any
+	// other locks.  This prevents a lock-order reversal when
+	// dtable_for_class is called from something holding the runtime lock while
+	// we're still holding the initialize lock.  We should ensure that we never
+	// acquire the runtime lock after acquiring the initialize lock.
+	LOCK_RUNTIME();
+
 	// Superclass +initialize might possibly send a message to this class, in
 	// which case this method would be called again.  See NSObject and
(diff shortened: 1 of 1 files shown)


dtable.c
```diff
@@ -824,26 +824,31 @@ PRIVATE void objc_send_initialize(id object)
 		objc_send_initialize((id)class->super_class);
 	}
 
+	// Lock the runtime while we're creating dtables and before we acquire any
+	// other locks.  This prevents a lock-order reversal when
+	// dtable_for_class is called from something holding the runtime lock while
+	// we're still holding the initialize lock.  We should ensure that we never
+	// acquire the runtime lock after acquiring the initialize lock.
+	LOCK_RUNTIME();
```
davidchisnall: Wouldn't LOCK_RUNTIME_FOR_SCOPE() work here?

ngrewe: I'm not sure. We want to drop the lock early before exiting the scope ([l. 876](https://github.com/ngrewe/libobjc2/blob/6b285936bed090d06f3bfc66d0886c12dc088753/dtable.c#L876)) to avoid that blocking unrelated runtime operations while +initialize runs for the class.

davidchisnall: Ah yes.  Looks fine then.

davidchisnall:
It would be nice to have a test for this, if it's possible to put one together (is it possible to force the threads into the state where it happens?)

ngrewe:
> is it possible to force the threads into the state where it happens?

I haven't found one. We've seen it in production exactly once and I could only reproduce by putting a lot of thread-specific breakpoints in the debugger and essentially manually scheduling them until thy deadlocked.

PR #22, opened 2016-10-27
Fix sarray2.h to build with gcc

ngrewe:
Hi David,

gcc and clang disagree on to deal with const expressions in static initialisations (bug report [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53091)), so the sparse array code currently doesn't compile with gcc. Making data_size and data_mask #defines when not compiling with clang helps.

### DIFF
--- a/sarray2.h
+++ b/sarray2.h
@@ -21,7 +21,7 @@
 #ifdef __clang__
 static const uint32_t data_size = 256;
 #else
-#define data_size 256
+#define data_size (uint32_t)256
 #endif
 /**
  * The mask used to access the elements in the data array in a sparse array
--- a/sarray2.h
+++ b/sarray2.h
@@ -18,12 +18,21 @@
  * The size of the data array.  The sparse array is a tree with this many
  * children at each node depth.
  */
+#ifdef __clang__
 static const uint32_t data_size = 256;
+#else
+#define data_size 256
+#endif
 /**
  * The mask used to access the elements in the data array in a sparse array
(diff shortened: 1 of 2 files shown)


ngrewe:
I've also added gcc to the travis build matrix to catch stuff like that. The tests don't build currently (which is kinda expected). I'm not sure whether I'm interested in sifting through them to see which ones we should be able to run with gcc…

sarray2.h
```diff
@@ -18,12 +18,21 @@
  * The size of the data array.  The sparse array is a tree with this many
  * children at each node depth.
  */
+#ifdef __clang__
 static const uint32_t data_size = 256;
+#else
+#define data_size 256
```
davidchisnall: Can we at least make this `(uint32_t)256`?

ngrewe: Definitely.

ngrewe:
> I don't really like this, as it's a work around for a gcc bug and is likely to mean that we carry around code supporting a buggy compiler

I don't have any particular interest in gcc as an Objective-C compiler either, but there is case to be made that libobjc2 still works as a drop-in replacement for the gcc runtime. 

That being said, I think it would be nice to deprecate gcc support in the 1.9.0, together with all the legacy runtime functions, so that we can eventually remove them?

davidchisnall:
I tried building with gcc a few months ago and almost all of the tests fail.  I think it's too broken to support.

PR #24, opened 2016-11-08
Test case for misaligned ivars

ngrewe:


Test/alignTest.m
```diff
@@ -52,8 +52,32 @@ - (void)check
...
+
 int main(void)
 {
 	[[Vector alloc] permute];
 	[[Foo new] check];
+
+	Ivar v_isa = class_getInstanceVariable([StringLikeTest class], "isa");
+	Ivar v_c_string = class_getInstanceVariable([StringLikeTest class], "c_string");
+	Ivar v_len = class_getInstanceVariable([StringLikeTest class], "len");
+	assert(ivar_getOffset(v_isa) < ivar_getOffset(v_c_string));
```
davidchisnall: Can you also assert that the offset of `isa` is 0 and the offset of `c_string` is `sizeof(Class)` please?

PR #33, opened 2017-06-01
Add objc-runtime.h to objc headers

theiostream:
Some applications still include the old objc-runtime.h header expecting
to obtain the declarations contained in objc/message.h and
objc/runtime.h nowadays. This commit adds this header to achieve
out-of-the-box compatibility with these applications.

(Context: this is one of the issues that arise when trying to build WebKit against GNUstep's libobjc2.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,7 @@ set(libobjc_HDRS
 	objc/objc-api.h
 	objc/objc-arc.h
 	objc/objc-auto.h
+	objc/objc-runtime.h
 	objc/objc.h
 	objc/runtime-deprecated.h
 	objc/runtime.h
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -14,17 +14,17 @@ set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fexceptions")
...
-set(libobjc_OBJC_SRCS 
+set(libobjc_OBJC_SRCS
 	NSBlocks.m
 	Protocol2.m
 	arc.m
 	associate.m
 	blocks_runtime.m
 	properties.m)
-set(libobjc_C_SRCS 
+set(libobjc_C_SRCS
```
davidchisnall: Please separate whitespace-only changes into a separate commit.

theiostream:
I amended the whitespace issue. Could you take a look again?

davidchisnall:
Merged.  Do you want to send a PR for the block headers as well?

PR #35, opened 2017-06-06
blocks_runtime: create aliases for type encoding getters

theiostream:
Add _Block_has_signature() and _Block_signature() alises for calls to
block_getType_np() for increased compatibility with other
implementations of libobjc2.

**Reasoning**: WebKit uses private libobjc APIs like `_Block_signature` to do what `block_getType_np` does, and requires them to be available.

Actually, adding these macros still wouldn't make that part of WebKit fully compatible with libobjc2, since WK actually expects `_Block_signature` to be an undeclared exported symbol in libobjc (it redeclares the function as an `extern` in its own code, which would fail compilation under the changes in this pull request). But reproducing these conditions in libobjc2 just wouldn't make any sense.

However, having this alias in libobjc2 still seems like it could serve to improve compatibility between libobjc2 and Apple's libobjc – although I don't know if the project has previously made a decision to ignore parity with private APIs. In this case, I'll understand if the pull request is rejected and will move these changes to my GNUstep-compatible fork of WebKit.

### DIFF
--- a/objc/blocks_runtime.h
+++ b/objc/blocks_runtime.h
@@ -2,6 +2,10 @@
 #pragma clang system_header
 #endif
 
+#ifndef __has_feature
+#	define __has_feature(x) 0
+#endif
+
 /*
  * Blocks Runtime
  */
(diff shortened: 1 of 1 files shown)


theiostream:
Any comments here?

After some thought I've come to consider that maybe these macros really don't help at all (since it'd still break compatibility with WebKit and would just clutter us with a bad way of implementing a private API). But I'm growing to consider the idea of creating these functions in libobjc2 and exporting them.

The only issue is that for compatibility purposes they would not be able to be declared (else we'd face redefinition errors).

davidchisnall:
This is pretty funny - the Apple folk objected a bit when I added type signatures to blocks, so I'm quite surprised to see that their code depends on it now!  If they're now exposing these as symbols, then I'd rather we switch to their API and leave the existing ones as compat wrappers.

davidchisnall:
The 1.9 and master branch now implement the two functions that WebKit wants.

PR #38, opened 2017-06-30
Add objc-class.h to objc headers

theiostream:
This one is pretty much equal to that other pull request for objc-runtime.h, but with a different header that just came up. :)


davidchisnall:
I'll merge this, though it would probably be better if we only exposed the APIs that the Apple runtime does from each of these headers, rather than exposing everything.

Issue #39, opened 2017-07-03
object_setInstanceVariable() -- Segfault


xFAR:
The current (1.8.1 and Git master) implementations of object_setInstanceVariable() trigger a Segault if the caller tries to set a non-existent ivar.  This can happen when a NIB is loaded.  A graceful exit with a NULL return value would seem to be better behavior. 

With 1.8.x I fix this issue with the following:

```
void object_setIvar(id object, Ivar ivar, id value)
{
	char *addr = (char*)object;
+	unsigned offset;
	
+	if (ivar && (offset = ivar_getOffset(ivar)))
+	    {
-	addr += ivar_getOffset(ivar);
+	    addr += offset;
	    *(id*)addr = value;
+	    }
}
```

The Git master code also blows up when determining the encoding type of a non-existent ivar:

```
Ivar object_setInstanceVariable(id obj, const char *name, void *value)
{
	Ivar ivar = class_getInstanceVariable(object_getClass(obj), name);
-->	if (ivar_getTypeEncoding(ivar)[0] == '@')

```

davidchisnall:
Please can you submit a pull request with this patch?

davidchisnall:
This was fixed by merging the pull request.

PR #41, opened 2017-07-19
1.8.1-FAR

xFAR:
A more robust fix would perform bounds checking of the Ivar's offset (s/b within a min/max for the class).  The object_getIvar() function should also probably return nil if passed a NULL Ivar (or a nil object). 

This patch does not address the second item mentioned in Issue #39 which affects the Master branch.

### DIFF
--- a/ivar.c
+++ b/ivar.c
@@ -149,8 +149,13 @@ PRIVATE void objc_compute_ivar_offsets(Class class)
 void object_setIvar(id object, Ivar ivar, id value)
 {
 	char *addr = (char*)object;
-	addr += ivar_getOffset(ivar);
-	*(id*)addr = value;
+	unsigned offset;
+	
+	if (ivar && (offset = ivar_getOffset(ivar)))
+		{
+		addr += offset;
+		*(id*)addr = value;
+		}
 }
 
 Ivar object_setInstanceVariable(id obj, const char *name, void *value)


ivar.c
```diff
@@ -149,8 +149,13 @@ PRIVATE void objc_compute_ivar_offsets(Class class)
 void object_setIvar(id object, Ivar ivar, id value)
 {
 	char *addr = (char*)object;
-	addr += ivar_getOffset(ivar);
-	*(id*)addr = value;
+	unsigned offset;
+	
+	if (ivar && (offset = ivar_getOffset(ivar)))
+		{
```
davidchisnall: Please fix indenting.

ivar.c
```diff
@@ -149,8 +149,13 @@ PRIVATE void objc_compute_ivar_offsets(Class class)
...
+		addr += offset;
+		*(id*)addr = value;
+		}
```
davidchisnall: Please fix indenting.

ivar.c
```diff
@@ -149,8 +149,13 @@ PRIVATE void objc_compute_ivar_offsets(Class class)
 void object_setIvar(id object, Ivar ivar, id value)
 {
 	char *addr = (char*)object;
-	addr += ivar_getOffset(ivar);
-	*(id*)addr = value;
+	unsigned offset;
+	
+	if (ivar && (offset = ivar_getOffset(ivar)))
```
davidchisnall: libobjc2 coding style follows the Étoilé guidelines and recommends avoiding implicit conversion from pointer to boolean, so please write this as `(ivar != NULL) && ...`.

davidchisnall:
The test failure looks like Travis is having problems, but please add a test case for this as well.

PR #45, opened 2017-11-18
Update blocks_runtime.h

zenny-chen:
Get Compatible with Apple LLVM.

### DIFF
--- a/blocks_runtime.m
+++ b/blocks_runtime.m
@@ -42,9 +42,9 @@ static void *_HeapBlockByRef = (void*)1;
 /**
  * Returns the Objective-C type encoding for the block.
  */
-const char *block_getType_np(void *b)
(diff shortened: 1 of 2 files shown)


zenny-chen:
Let Blocks API compatible with Apple LLVM.

davidchisnall:
The middle commit was passing CI tests, but this one doesn't seem to build.

danilaml:
@davidchisnall can you also merge https://github.com/gnustep/libs-base/pull/13 ? Currently you'll get incompatible headers when building from latest source since they'll contain different type signatures.

davidchisnall:
Merged, though -base shouldn't actually need that - a better fix would be to make the -base declarations conditional on `!__has_include(<objc/blocks_runtime.h>)`

PR #48, opened 2017-12-15
clang x86_64-w64-mingw32 changes

Sabine-Schnabeltier:
clang target x86_64-w64-mingw32 uses LLP64 model so defined offsets for this model to asmconstants.h and altered unwind-itanium.h to directly include definitions from the compiler's unwind.h

Other changes add prototype for mprotect and implementation for valloc and change static keyword to inline to remove clang error.

These changes eliminate compile errors; there are still the following link errors:

```
[ 17%] Linking C shared library objc2.dll
CMakeFiles/objc.dir/properties.m.obj:(.rdata+0xa2): multiple definition of `.objc_sel_namecopy'
CMakeFiles/objc.dir/associate.m.obj:(.rdata+0x43): first defined here
CMakeFiles/objc.dir/Protocol2.m.obj:(.text+0x5d): undefined reference to `__imp___objc_exec_class'
CMakeFiles/objc.dir/arc.m.obj:(.text+0x1d8d): undefined reference to `__imp___objc_exec_class'
CMakeFiles/objc.dir/associate.m.obj:(.text+0xfad): undefined reference to `__imp___objc_exec_class'
CMakeFiles/objc.dir/properties.m.obj:(.text+0x1acd): undefined reference to `__imp___objc_exec_class'
```

### DIFF
--- a/spinlock.h
+++ b/spinlock.h
@@ -1,6 +1,6 @@
-#ifdef __MINGW32__
+#if defined(__MINGW32__) && !defined(__MINGW64__)
 #include <windows.h>
-inline unsigned sleep(unsigned seconds)
+static unsigned sleep(unsigned seconds)
 {
 	Sleep(seconds*1000);
 	return 0;
--- a/spinlock.h
+++ b/spinlock.h
@@ -1,6 +1,6 @@
 #ifdef __MINGW32__
 #include <windows.h>
-static unsigned sleep(unsigned seconds)
+inline unsigned sleep(unsigned seconds)
 {
 	Sleep(seconds*1000);
 	return 0;
--- a/unwind-itanium.h
+++ b/unwind-itanium.h
@@ -33,6 +33,11 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 extern "C" {
 #endif
 
+/* definitions below do not handle LLP64, so use compiler definitions */
+#ifdef __MINGW32__
+#include <unwind.h>
+#else
+
 /* Minimal interface as per C++ ABI draft standard:
 
 	http://www.codesourcery.com/cxx-abi/abi-eh.html */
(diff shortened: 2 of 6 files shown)


spinlock.h
```diff
@@ -1,6 +1,6 @@
 #ifdef __MINGW32__
 #include <windows.h>
-static unsigned sleep(unsigned seconds)
+inline unsigned sleep(unsigned seconds)
```
davidchisnall: Why this change?  We don't want to be exporting a symbol called `sleep` from the library.

unwind-itanium.h
```diff
@@ -33,6 +33,11 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 extern "C" {
 #endif
 
+/* definitions below do not handle LLP64, so use compiler definitions */
+#ifdef __MINGW32__
+#include <unwind.h>
+#else
```
davidchisnall: Did you test whether this actually works?  I don't know how MinGW handles exceptions - does it use UNIX-style DWARF unwinding, or does it use the Win64 EH ABI?  If it's the latter, then we should get the patches from Microsoft and default to that.

spinlock.h
Sabine-Schnabeltier: Good point, hadn't thought of that!

I was trying to work around this clang error:

`In file included from libobjc2/properties.m:10:`
`libobjc2/spinlock.h:3:17: error: static declaration of 'sleep' follows non-static declaration`
`static unsigned sleep(unsigned seconds)`
`                ^`
`/opt/local/x86_64-w64-mingw32/include/unistd.h:46:22: note: previous declaration is here`
`unsigned int __cdecl sleep (unsigned int);`

Could instead replace function implementation with:

`#define sleep(seconds)  Sleep(seconds*1000)`

What do you think?

unwind-itanium.h
Sabine-Schnabeltier: Haven't been able to check this change because of the link errors listed above. Any idea how I can work around that?

Sabine-Schnabeltier:
I am closing this request for now to fix and consolidate the changes. Will re-open after more testing.

zhaozg:
Take care about this.

PR #50, opened 2017-12-22
Minimal reproducing test case

dlobron:
This creates a test number 47 in the unit test suite, which fails with an exception.  The exception reproduces the bug.

### DIFF
--- a/Test/CMakeLists.txt
+++ b/Test/CMakeLists.txt
@@ -5,7 +5,7 @@
 
 # List of single-file tests.
 set(TESTS
-	alignTest.m
+	#alignTest.m
 	AllocatePair.m
 	AssociatedObject.m
 	AssociatedObject2.m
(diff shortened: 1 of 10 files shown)


davidchisnall:
This seems to refer to a `ModTest.{m,h}`, which are not part of the pull request.

dlobron:
Ah, drat, I forgot to include them in the PR. I’m away from computer for this week, but I will add them as soon as I can. The files essentially duplicate the Test class from Test.{m|h}, but with interface and implementation separated to allow inclusion in multiple compilation units.

David

davidchisnall:
Never mind.  I've managed to delete about 90% of the code and have a minimal test case.

dlobron:
Thanks, David. Apologies for the extra code: it was left from earlier experiments. 

David

PR #54, opened 2018-02-22
Wrap some headers in extern "C" for C++ compatibility

DHowett-MSFT:


DHowett-MSFT:
I'm not entirely sold on this being the right choice; if we have `sys/cdefs` available to us, that might be more interesting/valid? `__BEGIN_DECLS` and friends are double-underscored, though, so perhaps we're not intended to use them.

PR #55, opened 2018-02-22
Use the old association policy to determine whether to release an object

DHowett-MSFT:
This commit has been updated to address @davidchisnall's comments in https://github.com/Microsoft/libobjc2/commit/df7f9061670cce994730d1720150adc


davidchisnall:
Would be nice to have a test for this too...

PR #56, opened 2018-02-22
teach class_setSuperclass about metaclasses, subclass lists, and dtables

DHowett-MSFT:


runtime.c
```diff
@@ -519,8 +521,44 @@ Class class_setSuperclass(Class cls, Class newSuper)
...
+	else
+	{
+		// newSuper is presumably a metaclass. Its isa will therefore be the appropriate root metaclass.
+		cls->isa = newSuper->isa;
+	}
+
+	// Make sure the superclass is initialized if we're initialized.
+	if (objc_test_class_flag(cls, objc_class_flag_initialized))
+	{
+		objc_send_initialize(newSuper);
```
davidchisnall: This is not safe.  `objc_send_initialize` expects to be called *without* the runtime lock being held and will both acquire and release it.  If it can't release the lock, then this can deadlock.

runtime.c
```diff
@@ -519,8 +521,44 @@ Class class_setSuperclass(Class cls, Class newSuper)
...
+		// Update the class's dtable to reflect its new superclass's dtable.
+		if (cls->dtable != uninstalled_dtable)
+		{
+			// we can't use objc_update_dtable_for_class here, as it doesn't take into account
```
davidchisnall: Why not simply set it to `uninstalled_dtable` and then let the runtime lazily initialise it the next time a message is sent to the class?  And presumably you also need to find all subclasses and do the same thing for their dtables, or they will still have incorrectly inherited methods?

runtime.c
```diff
@@ -519,8 +521,44 @@ Class class_setSuperclass(Class cls, Class newSuper)
...
+			// superclasses. It only walks downward.
+			free_dtable(cls->dtable);
+			cls->dtable = uninstalled_dtable;
+			cls->dtable = create_dtable_for_class(cls, uninstalled_dtable);
```
davidchisnall: Do you not also need to update (or, at least, invalidate) the metaclass's dtable?

runtime.c
```diff
@@ -519,8 +521,44 @@ Class class_setSuperclass(Class cls, Class newSuper)
...
+	// Make sure the superclass is initialized if we're initialized.
+	if (objc_test_class_flag(cls, objc_class_flag_initialized))
+	{
+		objc_send_initialize(newSuper);
+		// Update the class's dtable to reflect its new superclass's dtable.
+		if (cls->dtable != uninstalled_dtable)
+		{
+			// we can't use objc_update_dtable_for_class here, as it doesn't take into account
+			// superclasses. It only walks downward.
+			free_dtable(cls->dtable);
```
davidchisnall: This doesn't look safe.  If this call races with a call to `objc_msgSend` on this object / class in another thread then we're deleting the dtable out from under the other thread.  I think that we need to use something that updates the dtable in place to avoid that.  The entries in the dtable should be safe to (relaxed consistency) atomically replace.

davidchisnall:
It would also be good to have some a test for this, checking that changing the superclass of a class calls the correct inherited instance and class methods in both the class that is changed and a subclass of it.

DHowett-MSFT:
Changes in the latest iteration:
* We now merge dtables inplace, only under the runtime lock. Additions/merges down from the new superclass first, deletions for things not in the new superclass second.
  * I just realized I should probably be updating slot versions when deleting slot entries to invalidate external caches
* There's a test now. It fails spectacularly on master, and equally spectacularly with the changes in the previous iteration.
* `class_setSuperclass` isn't really functional for metaclasses directly; should anyone really be doing that?
* `+initialize` now happens outside of the runtime lock, and its status is checked under the initialize lock.
* There are a few points at which competing threads can observe an unusual state:
  * new superclass must be initialized, but has not yet been initialized
  * new superclass has been initialized, but dtable is not yet updated

I'm not sure how critical those two cases are. The second is a strict improvement over "`objc_msgSend` tearing through a `free()`'d dtable".

Tests were run on:
* Linux TDD x64 (modern/legacy dispatch)
* Windows TDD x32 (legacy dispatch)
* Linux TDD LOW_MEMORY x64 (legacy dispatch)
* Linux NO_TDD x64 (modern/legacy dispatch)

PR #59, opened 2018-02-22
win32: use native threading APIs instead of pthreads

DHowett-MSFT:
* use fiber local storage if NO_PTHREADS is defined
* use critical sections instead of mutexes

### DIFF
--- a/arc.m
+++ b/arc.m
@@ -12,9 +12,44 @@
 #import "objc/objc-arc.h"
 #import "objc/blocks_runtime.h"
 
-#ifndef NO_PTHREADS
-#include <pthread.h>
-pthread_key_t ARCThreadKey;
+#if defined(_WIN32)
+// We're using the Fiber-Local Storage APIs on Windows
+// because the TLS APIs won't pass app certification.
+// Additionally, the FLS API surface is 1:1 mapped to
+// the TLS API surface when fibers are not in use.
+#	include "safewindows.h"
+#	define arc_tls_store FlsSetValue
+#	define arc_tls_load FlsGetValue
+#	define TLS_CALLBACK(name) void WINAPI name
+
+typedef DWORD arc_tls_key_t;
+typedef void WINAPI(*arc_cleanup_function_t)(void*);
+static inline arc_tls_key_t arc_tls_key_create(arc_cleanup_function_t cleanupFunction)
+{
+	return FlsAlloc(cleanupFunction);
+}
+
+#else // if defined(_WIN32)
+
+#	ifndef NO_PTHREADS
+#		include <pthread.h>
+#		define arc_tls_store pthread_setspecific
+#		define arc_tls_load pthread_getspecific
+#		define TLS_CALLBACK(name) void name
+
+typedef pthread_key_t arc_tls_key_t;
+typedef void (*arc_cleanup_function_t)(void*);
+static inline arc_tls_key_t arc_tls_key_create(arc_cleanup_function_t cleanupFunction)
+{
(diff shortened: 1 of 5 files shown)


DHowett-MSFT:
Moderately contentious:

* fiber-local storage _is_ thread-local storage on all versions of Windows that matter
* the thread-local storage APIs (`Tls*`) are not available to store applications
* `NO_PTHREADS` is not the right thing to key off here
* We switched to critical sections because they're cheaper than mutexes. Mutexes can cross process boundaries and are a kernel affordance, and that incurs a round trip through kernel mode to manipulate them. CS are entirely in-process and usermode.

lock.h
```diff
@@ -10,11 +10,11 @@
...
 #	include <windows.h>
 #undef BOOL
-typedef HANDLE mutex_t;
-#	define INIT_LOCK(x) x = CreateMutex(NULL, FALSE, NULL)
-#	define LOCK(x) WaitForSingleObject(*x, INFINITE)
-#	define UNLOCK(x) ReleaseMutex(*x)
-#	define DESTROY_LOCK(x) CloseHandle(*x)
+typedef CRITICAL_SECTION mutex_t;
+#	define INIT_LOCK(x) InitializeCriticalSection(&(x))
+#	define LOCK(x) EnterCriticalSection(x)
```
DHowett-MSFT: critical sections can time out after about 10 minutes. this was deemed an acceptable cost.

arc.m
```diff
@@ -14,9 +14,36 @@
...
+#define arc_tls_load pthread_getspecific
+#define TLS_CALLBACK(name) void name
+
+static inline pthread_key_t arc_tls_key_create(void(*cleanupFunction)(void*))
+{
+	pthread_key_t key;
+	pthread_key_create(&key, cleanupFunction);
+	return key;
+}
+#else // if defined(NO_PTHREADS)
```
davidchisnall: This should be in some kind of `#ifdef` Windows thing - Windows isn't the only platform with no pthreads.  I suspect that for Windows this actually shouldn't be in the `#else` bit at all: even if we have Pthreads on Windows via the MinGW pthreads implementation, we'd probably prefer to use the native Windows APIs.

arc.m
```diff
@@ -60,17 +87,13 @@ - (void)release;
 
 static inline struct arc_tls* getARCThreadData(void)
 {
-#ifdef NO_PTHREADS
-	return NULL;
-#else
-	struct arc_tls *tls = pthread_getspecific(ARCThreadKey);
+	struct arc_tls *tls = arc_tls_load(ARCThreadKey);
```
davidchisnall: We still need to keep the `NO_PTHREADS` case for platforms that both are not Windows and don't support pthreads.

lock.h
davidchisnall: No objection to this, but I wondered if you'd done any profiling of Mutex vs CriticalSection vs SRW lock for workloads that you care about.

DHowett-MSFT: The data there is, unfortunately, lost to the ages. @bviglietta made the switch for us ~2 years ago.

DHowett-MSFT:
We now prefer Fls on Windows (even MinGW). Fls is available down to Vista/Server 2003.

DHowett-MSFT:
Fixed an issue in `eh_personality.c` that came up building with NO_PTHREADS. It looks like ARC-compatible autorelease doesn't work at all with pthreads disabled (I'll file an issue on this).

Additionally, some Windows defines were preventing us building properly outside WinObjC; those are now mostly fixed.

PR #64, opened 2018-03-04
Fix some link-stage errors when compiling with Cygwin

CNAmira:
Some errors still here such as
```
CMakeFiles/objc.dir/arc.m.o:(.text+0x2553): undefined reference to `_imp____objc_exec_class'
```

I found that those errors only occur on objc objects, I think it might be some bugs in clang or ld.

### DIFF
--- a/common.S
+++ b/common.S
@@ -1,4 +1,4 @@
-#if (defined(_WIN32) && defined(__i386__)) || defined(_CYGWIN) || defined(__APPLE__)
+#if ((defined(_WIN32) || defined(__CYGWIN__)) && defined(__i386__)) || defined(__APPLE__)
 #define CDECL(symbol) _##symbol
 #else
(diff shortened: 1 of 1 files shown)


davidchisnall:
@DHowett-MSFT Does this look right to you, or is the `_WIN32` define wrong?

DHowett-MSFT:
I think adding _CYGWIN is correct: _WIN32 is compiler-provided for MSVC, GCC and Clang for Windows non-POSIX builds, and Cygwin doesn’t provide it so as to not confuse cross-platform code.

If there is a compiler-provided define that covers more of the spectrum of windows targets such that we don’t need to pass -DWINDOWS, that’d definitely be an improvement!

CNAmira:
_CYGWIN is not defined internally.
symbols of cdecl functions don't have '_' prefix either.

Issue #66, opened 2018-04-03
cxa_allocate_exception is not compatible with CLang 6


kwhat:
Hi,

It looks like there is an issue with clang 6 compatibility related to cxa_allocate_exception and related stuff. I have patched the issue against 1.8 as well as a bunch of warnings.  I think the warnings were taken care of in master but I am not sure if the cxa_allocate_exception was addressed.  I didn't see any reported bugs related to this issue so I have opened this.

__clang_cxa_allocate_exception.patch__ 
```
diff -Naur libobjc2-1.8.1/CMakeLists.txt libobjc2-1.8.1.new/CMakeLists.txt
--- libobjc2-1.8.1/CMakeLists.txt       2015-08-07 04:33:40.000000000 -0700
+++ libobjc2-1.8.1.new/CMakeLists.txt   2018-04-03 13:55:39.233515034 -0700
@@ -7,6 +7,7 @@
 set(CMAKE_C_FLAGS_DEBUG "-g -O0 -fno-inline ${CMAKE_C_FLAGS_DEBUG}")
 set(CMAKE_C_FLAGS_RELEASE "-O3 ${CMAKE_C_FLAGS_RELEASE}")
 set(CMAKE_C_FLAGS "-std=gnu99 ${CMAKE_C_FLAGS}")
+set(CMAKE_CXX_FLAGS "-std=c++11 -stdlib=libc++ ${CMAKE_CXX_FLAGS}")
 
 set(libobjc_VERSION 4.6)
 
@@ -14,7 +15,7 @@
 # Build configuration
 add_definitions( -DGNUSTEP -D__OBJC_RUNTIME_INTERNAL__=1)
 # Probably not needed anymore?
-add_definitions( -D_XOPEN_SOURCE=700 -D__BSD_VISIBLE=1 -D_BSD_SOURCE=1)
+add_definitions( -D_XOPEN_SOURCE=700 -D_DEFAULT_SOURCE=1)
 
 set(libobjc_ASM_SRCS 
        block_trampolines.S
diff -Naur libobjc2-1.8.1/Makefile libobjc2-1.8.1.new/Makefile
--- libobjc2-1.8.1/Makefile     2015-08-07 04:33:40.000000000 -0700
+++ libobjc2-1.8.1.new/Makefile 2018-04-03 13:52:44.593214733 -0700
@@ -18,7 +18,7 @@
 #CFLAGS += -Wno-deprecated-objc-isa-usage
[...]

PR #67, opened 2018-04-03
Fixing issue with cxa_allocate_exception #66

kwhat:


kwhat:
I believe this was fixed by #68

PR #70, opened 2018-04-10
Let go of the runtime lock properly in class_setSuperclass

DHowett-MSFT:
This will almost certainly cause a deadlock if you use `class_setSuperclass` incorrectly, which WinObjC apparently does!

### DIFF
--- a/runtime.c
+++ b/runtime.c
@@ -526,7 +526,10 @@ Class class_setSuperclass(Class cls, Class newSuper)
 
 	LOCK_RUNTIME();
 
-	if (cls->super_class == newSuper) { return newSuper; }
+	if (cls->super_class == newSuper) {
+		UNLOCK_RUNTIME();
+		return newSuper;
+	}
 
 	safe_remove_from_subclass_list(cls);
 	objc_resolve_class(newSuper);


davidchisnall:
I've refactored this a bit to use the LOCK_FOR_SCOPE macro.

DHowett:
That’s a lot easier to follow. Thanks!

davidchisnall:
By the way, your test for this function is failing for me locally, but not in Travis.

davidchisnall:
It appears that the test failure was caused by not providing space for the isa pointer.  I've now fixed that and added a check to make sure that we don't create objects with size 0.  This probably should be a check that we don't create objects smaller than `sizeof(id)`.

I wonder if we also need some checks in setSuperclass that the new and old superclasses have the same size.  I'm beginning to see why Apple deprecated this function in 10.5.  What are you using it for?

DHowett-MSFT:
Thanks for all the fixes. Sorry about the broken test. :frowning:

> What are you using it for?

Since WinObjC implements some CoreFoundation types with support for toll-free bridging, we needed the ability to emit classes in CoreFoundation whose hierarchies were replaced once Foundation was loaded (if it was going to be loaded). All our bridging types under Foundation inherit from `_NSCFTemporaryRootClass` (size=`sizeof(id)`) and that's replaced with `NSObject` later.

I know Apple in particular keeps `NSObject` in objc4, but that's a cost we couldn't countenance.

PR #88, opened 2019-03-01
Fix dtable rebase and add a test for a missing case

DHowett-MSFT:
The dtable rebasing rework in b81df02b broke the inheritance of
superclass methods from a superclass that had already been initialized.


DHowett-MSFT:
(I ran into this testing Clang 9 with WinObjC.)

PR #89, opened 2019-03-01
Scan ABIv2 tables pointer-by-pointer to skip intra-section padding

DHowett-MSFT:
On Windows with incremental linking enabled, structures laid out in the
.objcrt sections may be padded on either side with alignment-sized sets
of null bytes. Since Clang sets the alignment for each ABI section to a
pointer's width, we can skip over those padding bytes by moving
uintptr_twise.

### DIFF
--- a/loader.c
+++ b/loader.c
@@ -207,26 +207,29 @@ OBJC_PUBLIC void __objc_load(struct objc_init *init)
 			break;
 	}
 	assert(init->version == 0);
-	assert((((uintptr_t)init->sel_end-(uintptr_t)init->sel_begin) % sizeof(*init->sel_begin)) == 0);
-	assert((((uintptr_t)init->cls_end-(uintptr_t)init->cls_begin) % sizeof(*init->cls_begin)) == 0);
-	assert((((uintptr_t)init->cat_end-(uintptr_t)init->cat_begin) % sizeof(*init->cat_begin)) == 0);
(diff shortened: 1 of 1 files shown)


DHowett-MSFT:
I've created this as a draft so that we can discuss it. I agree that this is best solved with a linker directive or by using another linker.

My main contention, which is primary centered around WinObjC (and therefore, we are willing to carry this patch on our own) is twofold:
* The release of a MSVC linker that supports this is months away, and will go out long after the first update of Visual Studio 2019.
* WinObjC isn't otherwise shipping a linker, and would not like to take on the burden of shipping the llvm linker in-package for WinObjC consumers to use.
  * (I don't have engineering time to dedicate to llvm-lld compatibility throughout the VS toolstack, unfortunately.)

This patch is fragile, but it tries to only rely on known invariants: that the first field of every runtime structure will never intentionally be `NULL`.

davidchisnall:
I think I'd rather keep this in the WinObjC version, where the ABI guarantees are weaker.  I'm not sure that using lld-link.exe instead of link.exe is that much of a problem (the LLVM toolchain files for Visual Studio will use lld-link and clang), but if it is then having this in WinObjC for a release or two and then removing it seems like a better option than upstreaming it.  I presume the directive will be silently ignored by linkers that don't support it, so we should be able to make clang emit it soon and then turn this into dead code as soon as the new linker is released.

PR #91, opened 2019-03-08
Replace valloc() with memalign() to support e.g. Android arm64

triplef:
valloc() no longer exists in some toolchains like Android arm64 (see [here](http://androidxref.com/5.0.0_r2/xref/bionic/libc/private/bionic_config.h#20)), but using `memalign(getpagesize(), len)` should be [identical](https://www.gnu.org/software/libc/manual/html_node/Aligned-Memory-Blocks.html#index-valloc).

Caveat: I haven’t been able to verify whether support for memalign() and getpagesize() is universal.

### DIFF
--- a/block_to_imp.c
+++ b/block_to_imp.c
@@ -56,11 +56,6 @@ void __clear_cache(void* start, void* end);
 #define PROT_EXEC  0x1
 #endif
 
-static void *valloc(size_t len)
-{
-	return VirtualAlloc(NULL, len, MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE);
-}
-
 static int mprotect(void *buffer, size_t len, int prot)
 {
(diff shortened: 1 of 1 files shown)


triplef:
Looks like this will need some more work. I found that memalign() is not available e.g. on macOS using the Xcode toolchain, so using posix_memalign() is probably the better choice.

davidchisnall:
memalign is a glibcism, but posix_memalign works pretty much everywhere (except Windows, but we're already special-casing Windows here).  Happy to merge a PR that uses posix_memalign.

triplef:
Thanks. I’ve updated the patch to use posix_memalign with getpagesize(). I think that should be the equivalent of valloc?

I ended up inlining the VirtualAlloc call on Windows, let me know if you prefer doing it another way.

PR #94, opened 2019-03-23
Use simple implementations for some weak ref table funnctions

jordo:
So I can't explain this one well...  However on the 1.9 branch, on our android devices we had a bunch of memory corruption issues in the arc weak reference table.  I managed to track a few of them down by compiling and running our software under Addresses Sanitizer on armv7-a arch devices.    

There *seemed* to be a hash table bug, and perhaps something related to these functions.  I swapped out the the implementations with what's suggested in the arc spec and stability improved.  I'm not sure what the performance impact was however.

### DIFF
--- a/arc.m
+++ b/arc.m
@@ -870,42 +870,14 @@ id objc_loadWeak(id* object)
 
 void objc_copyWeak(id *dest, id *src)
 {
-	// Don't retain or release.  While the weak ref lock is held, we know that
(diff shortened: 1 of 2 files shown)


jordo:
I added the text relocation fix as well to msgSend on arm32.  I'm not 100% sure but it may only work on ELF assemblers..

PR #95, opened 2019-03-24
Use simple implementations for some weak ref functions

jordo:
So I can't explain this one well... However on the 1.9 branch, on our android devices we had a bunch of memory corruption issues in the arc weak reference table. I managed to track a few of them down by compiling and running our software under Addresses Sanitizer on armv7-a arch devices.

There seemed to be a hash table bug, and perhaps something related to these functions. I swapped out the the implementations with what's suggested in the arc spec and stability improved. I'm not sure what the performance impact was however.

Feel free to ignore...

### DIFF
--- a/arc.m
+++ b/arc.m
@@ -870,42 +870,14 @@ id objc_loadWeak(id* object)
 
 void objc_copyWeak(id *dest, id *src)
 {
-	// Don't retain or release.  While the weak ref lock is held, we know that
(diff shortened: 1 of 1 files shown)


davidchisnall:
I don't think that these changes are correct, but looking at the existing code I don't think it's correct in the presence of concurrent mutation either.  Do you have a test that has two threads trying to update the same __weak variable that fails?

PR #102, opened 2019-05-16
Add support for atomic type qualifiers.

### LINKED ISSUE #101: Crash in alignof_type() when using atomic ivars
Adding an atomic ivar member to a class will lead to a crash at runtime:

```objc
#include <stdatomic.h>

@interface Test
{
	_Atomic(bool) test; // bool or any other atomic type
}
@end
```

The backtrace looks like this:

```
(lldb) bt
* thread #1, name = 'helloobjectivec', stop reason = signal SIGABRT
  * frame #0: 0xf0c7cb39 [vdso]`__kernel_vsyscall + 9
    frame #1: 0xee921df9 libc.so`syscall + 41
    frame #2: 0xee924ed4 libc.so`abort + 116
    frame #3: 0xd0e503cb libobjc.so`alignof_type(type="AB", align=0xffce5fd4) at encoding2.c:358
    frame #4: 0xd0e4fbdb libobjc.so`objc_alignof_type(type="AB") at encoding2.c:372
    frame #5: 0xd0e6168c libobjc.so`upgradeIvarList(cls=0xd0f786a8) at legacy.c:128
    frame #6: 0xd0e61288 libobjc.so`objc_upgrade_class(oldClass=0xd0f786a8) at legacy.c:333
    frame #7: 0xd0e524fa libobjc.so`__objc_exec_class(module=0xd0f78884) at loader.c:366
    frame #8: 0xd0f730a0 libnative-lib.so`.objc_load_function + 32
    frame #9: 0xf0ca0961 linker`__dl__ZL10call_arrayIPFviPPcS1_EEvPKcPT_jbS5_ + 305
```

It looks like `alignof_type()` doesn’t implement a case for atomic types (`A`).

This is using the [GNUstep Android toolchain](https://github.com/gnustep/tools-android) with clang from the Android NDK, and I was able to reproduce this with all supported ABIs. Please let me know if I can provide any further info. Thanks!

triplef:
Fixes #101.

### DIFF
--- a/Test/PropertyIntrospectionTest2_arc.m
+++ b/Test/PropertyIntrospectionTest2_arc.m
@@ -510,7 +510,9 @@ void structDefault2Setter(id self, SEL _cmd, struct YorkshireTeaStruct value) {
 
 int main(void)
 {
+#ifdef NEW_ABI
 	testProperty("atomicBoolDefault", "TAB,VatomicBoolDefault", ATTRS(ATTR("T", "AB"), ATTR("V", "atomicBoolDefault")));
+#endif
 	testProperty("charDefault", "Tc,VcharDefault", ATTRS(ATTR("T", "c"), ATTR("V", "charDefault")));
 	testProperty("doubleDefault", "Td,VdoubleDefault", ATTRS(ATTR("T", "d"), ATTR("V", "doubleDefault")));
 	testProperty("enumDefault", "Ti,VenumDefault", ATTRS(ATTR("T", "i"), ATTR("V", "enumDefault")));
@@ -601,7 +603,9 @@ int main(void)
                                                                                                           ATTR("S", "setDynamicGetterSetter:")));
 
     Protocol *testProto = objc_getProtocol("ProtocolTest");
+#ifdef NEW_ABI
     testPropertyForProtocol(testProto, "atomicBoolDefault", "TAB", ATTRS(ATTR("T", "AB")));
+#endif
     testPropertyForProtocol(testProto, "charDefault", "Tc", ATTRS(ATTR("T", "c")));
 	testPropertyForProtocol(testProto, "doubleDefault", "Td", ATTRS(ATTR("T", "d")));
 	testPropertyForProtocol(testProto, "enumDefault", "Ti", ATTRS(ATTR("T", "i")));
--- a/Test/PropertyIntrospectionTest2_arc.m
+++ b/Test/PropertyIntrospectionTest2_arc.m
@@ -3,6 +3,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <assert.h>
+#include <stdatomic.h>
 
 #pragma GCC diagnostic ignored "-Wobjc-property-no-attribute"
 
@@ -36,6 +37,7 @@ __attribute__((objc_root_class))
 {
 @public
 	Class isa;
+	atomic_bool atomicBoolDefault;
 	signed char charDefault;
 	double doubleDefault;
 	enum FooManChu enumDefault;
@@ -68,6 +70,7 @@ __attribute__((objc_root_class))
 	__weak id idReadonlyWeakNonatomic;
 	id _idOther;
 }
+@property atomic_bool atomicBoolDefault;
 @property signed char charDefault;
 @property double doubleDefault;
 @property enum FooManChu enumDefault;
@@ -111,6 +114,7 @@ __attribute__((objc_root_class))
 
 
 @implementation PropertyTest
+@synthesize atomicBoolDefault;
 @synthesize charDefault;
 @synthesize doubleDefault;
 @synthesize enumDefault;
@@ -147,6 +151,7 @@ __attribute__((objc_root_class))
 @end
 
 @protocol ProtocolTest
+@property atomic_bool atomicBoolDefault;
 @property signed char charDefault;
 @property double doubleDefault;
 @property enum FooManChu enumDefault;
@@ -187,6 +192,7 @@ __attribute__((objc_root_class))
 @interface PropertyProtocolTest <ProtocolTest>
 {
 	Class isa;
+	atomic_bool atomicBoolDefault;
 	signed char charDefault;
 	double doubleDefault;
 	enum FooManChu enumDefault;
@@ -222,6 +228,7 @@ __attribute__((objc_root_class))
 @end
 
 @implementation PropertyProtocolTest
+@synthesize atomicBoolDefault;
 @synthesize charDefault;
 @synthesize doubleDefault;
 @synthesize enumDefault;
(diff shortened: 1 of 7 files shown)


davidchisnall:
Looks good (no idea why Linux CI is unhappy - it appears to be an issue with the Azure hosted Linux agents).  Please can you add a test?

triplef:
I’ve added a test for atomic ivars (passing), and also one for properties with atomic types which is currently failing like this:

```
Assertion failed: (attrsCount == size), function testPropertyForProperty_alt, file PropertyIntrospectionTest2_arc.m, line 295.
```

It seems to me that property_copyAttributeList() requires modifications to support atomic types, but I can’t quite figure how it’s supposed to work (e.g. why does [this loop](https://github.com/gnustep/libobjc2/blob/master/properties.m#L389) start at `strlen(types)+1`?).

`property_getAttributes()` returns "TAB" for the "atomicBoolDefault" property I added, but `property_copyAttributeList()` returns two properties "T":"B", and "D":"" which seems incorrect.

I’d appreciate your help to fix this.

davidchisnall:
That really needs a comment (and not the one that's there already, because that stopped being true as part of the 2.0 ABI refactoring).  The property encoding string is T{Objective-C type encoding of the property value}{other stuff}.  The `strlen(types)+1` is skipping the T and the encoding of the type.

With the 2.0 (and 1.8ish or later version of the 1.x) ABI, the full attribute encoding is generated by the compiler, so this should be tproperty_getAttributesreated as canonical.  The `"TAB"` comes from there (T[ype] is A[tomic] B[oolean]). I'd expect `property_copyAttributeList` for this to just return `{ "T" : "AB" }`.  

I suspect the problem is that, for atomic types, the type of the property is `_Atomic(T)`, the type of the accessor is `T` and is therefore one character smaller.  I think this may be a clang bug, because the types field isn't intended to be the type of the accessor.  We can work around this in the runtime by seeing if the character at `strlen(types)+1` is a comma and, if not, fix up types by copying the substring from the attributes string.  

The `"D"` value feels like it's a different bug - can to single-step through that function and see where `count` is being incremented to give us the nonsense value?

triplef:
I think I found the issue in `property_copyAttributeList`: the default case in the switch statement was (I think) incorrectly incrementing the counter in addition to it being incremented by the for loop. This caused it to skip over the string end in this case where `types = "B"` and `attributes = "TAB"`, as the start of the loop was not the end of the attributes string but its last character.

However `property_copyAttributeList` will still return `{"T":"B"}` instead of `{"T":"AB"}`, I think due to `A` being skipped by `objc_skip_type_qualifiers`. I can’t say if this is correct/expected behavior.

triplef:
Hm the PropertyIntrospectionTest2_arc test passes for me on macOS using clang from the latest Xcode 10.2. Any idea why it might fail on the CI?

davidchisnall:
The test also fails for me locally.  It looks as if it's because you fixed the bug and not the test (maybe you didn't commit the test?).  The test is looking for `{ "T" : "B" }` but it's finding `{ "T" : "AB"}` (which is correct).

Test/PropertyIntrospectionTest2_arc.m
```diff
@@ -503,6 +510,7 @@ void structDefault2Setter(id self, SEL _cmd, struct YorkshireTeaStruct value) {
 
 int main(void)
 {
+	testProperty("atomicBoolDefault", "TAB,VatomicBoolDefault", ATTRS(ATTR("T", "B"), ATTR("V", "atomicBoolDefault")));
```
davidchisnall: Should be "AB" not "B".

Test/PropertyIntrospectionTest2_arc.m
```diff
@@ -593,6 +601,7 @@ int main(void)
                                                                                                           ATTR("S", "setDynamicGetterSetter:")));
 
     Protocol *testProto = objc_getProtocol("ProtocolTest");
+    testPropertyForProtocol(testProto, "atomicBoolDefault", "TAB", ATTRS(ATTR("T", "B")));
```
davidchisnall: I think this should also be "AB".

Test/ivar_atomic.m
```diff
@@ -0,0 +1,24 @@
...
+{
+	atomic_bool atomicBool;
+}
+@end
+
+@implementation Dummy
+- (void)test
+{
+  int value = 1;
+  object_setIvar(self, class_getInstanceVariable(object_getClass(self), "atomicBool"), (__bridge id)(void*)(intptr_t)value);
```
davidchisnall: This is interesting.  It's probably fine to do a non-atomic store of this, but if you do `_Atomic(struct foo)`, where `struct foo` is a large value, then we should be calling into the relevant libatomic functions for atomic operations to make sure that we acquire the correct locks.  I doubt Apple's runtime does this correctly, but we should probably open an issue to track it.

triplef:
I’ve updated the test to expect `{ "T" : "AB"}` in both cases, although this is not what it finds for me locally (it finds `{ "T" : "B"}`. I’m curious though if this will make them pass on the CI.

Test/ivar_atomic.m
triplef: Yeah I realize that this doesn’t provide atomicity, it was merely meant as a test for the ivar bug (strictly speaking just compiling the class would have been enough to test for #101). Let me know if I should change the test in any way.

triplef:
Looks like only the legacy variants of the test failed. Is this expected and what would be a way to handle that?

davidchisnall:
You can use `#if __OBJC_GNUSTEP_RUNTIME_ABI__ >= 20` in the tests to check for the new ABI.  It looks as if there's enough metadata generated by clang for this to work though, so I'd take a look at what's going on in `legacy.c` to upgrade property structures to the new ABI to see if there's a bug there.  It looks as if it's getting the type from the getter and not from the attribute string.

triplef:
Thanks, I’ve disabled the tests on legacy ABIs and the tests are passing on the CI (and also on my machine now, I guess I was using the old ABI).

TBH I’m not sure I have the context to really dig into this further, but please let me know if there are any further changes you’d like to see.

triplef:
If it makes sense I could also do the following:

* Reduce this PR to just fix atomic ivars (i.e. [this change in `objc_skip_type_qualifiers`](https://github.com/gnustep/libobjc2/pull/102/commits/e5dc1c55b5e4990f176d566ce736926c3e25e087#diff-83432cf2188db01f28aa18b52883ba51R24) + [the test](https://github.com/gnustep/libobjc2/pull/102/commits/e48efee1371bbd837bebc76df70d3c5a41a3e385))
* Optionally create a new draft PR for the other changes for atomic property and method parameter support so it can be looked at in more detail

This should unblock the fix for the probably more common use case of atomic ivars (which is also what we are running into) from the changes for atomic types in properties and method signatures (which I’m guessing are very rarely used).

davidchisnall:
I'd like to fix the legacy ABI issue as well.  I'll try to take a look in the next day or two.  I suspect that there's an almost identical bug to the one that you've fixed in the legacy upgrade path.

triplef:
Ok great, thanks!

davidchisnall:
I've merged this and fixed the legacy path.  The test now passes with both the old and new ABIs.  See: 375018a933d197d12bed7b937480cf912280fd13

triplef:
Great, thank you David!

PR #106, opened 2019-05-20
Fixed warning in imp_implementationWithBlock.

triplef:
```
[8/372] Building C object CMakeFiles/objc.dir/block_to_imp.c.o
../block_to_imp.c:186:23: warning: incompatible pointer types initializing 'struct Block_layout *' with an expression of type 'id' (aka 'struct objc_object *') [-Wincompatible-pointer-types]
        struct Block_layout *b = block;
                             ^   ~~~~~
```

Was caused by switching the method signature to use "id" for the block parameter.

### DIFF
--- a/block_to_imp.c
+++ b/block_to_imp.c
@@ -183,7 +183,7 @@ static struct trampoline_set *trampolines;
 
 IMP imp_implementationWithBlock(id block)
 {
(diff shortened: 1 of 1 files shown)


davidchisnall:
We should probably make the CI pass -Werror to the compiler...

PR #113, opened 2019-06-14
Pass on Android toolchain flags for CXX runtime check.

triplef:
When cross-compiling for Android, CMake requires custom flags that are used in the [CMake toolchain file](https://android.googlesource.com/platform/ndk/+/master/build/cmake/android.toolchain.cmake) provided by the NDK (e.g. `ANDROID_ABI`, `ANDROID_NDK`, `ANDROID_PLATFORM`).

For the C++ interop test, these custom flags don’t seem to get passed on when running `try_compile()` on the `test_cxx_runtime` binary.

This patch fixes the issue, but I’m not sure if that’s the right / best solution and whether that would cause issues on other platforms. Is there a better way to specify specific flags that should get passed on if available?


CMakeLists.txt
```diff
@@ -207,7 +207,7 @@ function(test_cxx CXX_RUNTIME_NAME IS_STDLIB)
 			"${CMAKE_BINARY_DIR}/CMake"
 			"${CMAKE_CURRENT_LIST_DIR}/CMake"
 			test_cxx_runtime
-			CMAKE_FLAGS "-DCXX_RUNTIME=${CXX_RUNTIME_LIB}")
+			CMAKE_FLAGS "-DCXX_RUNTIME=${CXX_RUNTIME_LIB}" "-DANDROID_ABI=${ANDROID_ABI}" "-DANDROID_NDK=${ANDROID_NDK}" "-DANDROID_PLATFORM=${ANDROID_PLATFORM}")
```
davidchisnall: Can we make this conditional on the target being android?

triplef: Makes total sense, I’ve made that change and also added all other flags documented here:
https://developer.android.com/ndk/guides/cmake#variables

PR #116, opened 2019-07-10
_Bool fix for cpp compilation

jordo:
_Bool isn't defined when compiling from (.mm) objc++ files.

### DIFF
--- a/objc/blocks_runtime.h
+++ b/objc/blocks_runtime.h
@@ -7,22 +7,19 @@
  * Blocks Runtime
  */
 #include "Availability.h"
+#include <stdbool.h>
 #ifdef __cplusplus
-#define OBJC_BOOL_TYPE bool
-#define OBJC_BLOCK_PTR_TYPE void*
 #define BLOCKS_EXPORT extern "C"
 #else
-#define OBJC_BOOL_TYPE _Bool
-#define OBJC_BLOCK_PTR_TYPE id
-#define BLOCKS_EXPORT extern 
+#define BLOCKS_EXPORT extern
 #endif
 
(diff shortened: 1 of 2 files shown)


davidchisnall:
What is the `#include <stdbool.h>` for?  This defines `bool` in C and should have no effect in C++ mode.  I think it would be cleaner to either `#include <stdbool.h>` in C mode and just use `bool` everywhere (may break old C code, but is probably fine) or define an `OBJC_BOOL` macro that is either `bool` or `_Bool` depending on whether we're in C or C++ mode.

jordo:
oops  I thought i have taken that out before the pull req, I can remove.  Just using bool does break C code in other libraries (dispatch) for example.    But ya, agreed OBJC_BOOL would be cleaner and I can guard it, as I wasn't sure if OBJC_BOOL could be defined on some other lib or something.

objc/blocks_runtime.h
```diff
@@ -2,35 +2,28 @@
...
 
+
 /*
  * Blocks Runtime
  */
 #include "Availability.h"
 #ifdef __cplusplus
-#include <stdbool.h>
+#define OBJC_BOOL_TYPE bool
+#define OBJC_BLOCK_PTR_TYPE void*
```
davidchisnall: Is this consistent with Apple?  I'm not sure why we ended up with `void*` in Objective-C++ and `id` in Objective-C.

jordo: OK.  maybe i'll refactor to be as consistent with apple as possible (they seem to include #include <stdbool.h> and just use bool as return type as well).  That may be best moving forward for compatibility with their header.

jordo: OK, should look better now.

PR #123, opened 2019-08-19
heap corruption in weakRef impl

jordo:
I don't think the isDeleted flag is getting checked when loading a weak pointer.  weakRefGetObj returns nil when flag is set, but loading the pointer doesn't care about it.

I'm not 100% sure this is correct, but our app has a whole bunch of issues without it.

### DIFF
--- a/arc.m
+++ b/arc.m
@@ -678,10 +678,6 @@ static BOOL loadWeakPointer(id *addr, id *obj, WeakRef **ref)
 	{
 		*ref = (WeakRef*)oldObj;
 		*obj = weakRefGetObj(*ref);
-		if(*obj == nil) {
-			*ref = NULL;
-			return NO;
-		}
 		return YES;
 	}
 	*ref = NULL;
--- a/arc.m
+++ b/arc.m
@@ -199,6 +199,8 @@ extern BOOL FastARCAutorelease;
 
 static BOOL useARCAutoreleasePool;
 
+static int weakref_class;
+
 static const long refcount_shift = 1;
 /**
  * We use the top bit of the reference count to indicate whether an object has
@@ -287,6 +289,10 @@ static inline id retain(id obj)
 	if (objc_test_class_flag(cls, objc_class_flag_fast_arc))
 	{
 		return objc_retain_fast_np(obj);
+	}
+	if ((Class)&weakref_class == cls) {
+		// compiler emit retain on weafRef
+		return obj;
 	}
 	return [obj retain];
 }
@@ -606,8 +612,6 @@ OBJC_PUBLIC id objc_storeStrong(id *addr, id value)
 ////////////////////////////////////////////////////////////////////////////////
 // Weak references
 ////////////////////////////////////////////////////////////////////////////////
-
-static int weakref_class;
 
 typedef struct objc_weak_ref
 {
--- a/arc.m
+++ b/arc.m
@@ -855,11 +855,6 @@ OBJC_PUBLIC id objc_loadWeakRetained(id* addr)
 	// will acquire the lock before attempting to deallocate)
 	if (obj == nil)
 	{
-		// If we've destroyed this weak ref, then make sure that we also deallocate the object.
-		if (weakRefRelease(ref))
-		{
-			*addr = nil;
-		}
 		return nil;
 	}
 	Class cls = classForObject(obj);
--- a/arc.m
+++ b/arc.m
@@ -678,6 +678,10 @@ static BOOL loadWeakPointer(id *addr, id *obj, WeakRef **ref)
 	{
 		*ref = (WeakRef*)oldObj;
 		*obj = weakRefGetObj(*ref);
+		if(*obj == nil) {
+			*ref = NULL;
+			return NO;
+		}
 		return YES;
 	}
 	*ref = NULL;
--- a/arc.m
+++ b/arc.m
@@ -759,6 +759,11 @@ WeakRef *incrementWeakRefCount(id obj)
 	else
 	{
 		assert(ref->object == obj);
+		// if the weakRef for obj isDeleted, it means obj has begun deallocation
+		if(ref->isDeleted) {
+			weakRefRelease(ref);
+			return NULL;
+		}
 		ref->weak_count++;
 	}
 	return ref;
@@ -767,39 +772,46 @@ WeakRef *incrementWeakRefCount(id obj)
 OBJC_PUBLIC id objc_storeWeak(id *addr, id obj)
 {
 	LOCK_FOR_SCOPE(&weakRefLock);
-	WeakRef *oldRef;
-	id old;
-	loadWeakPointer(addr, &old, &oldRef);
-	// If the old and new values are the same, then we don't need to do anything.
-	if (old == obj)
+	BOOL assignNoAndUnregister = NO;
+	if(obj == nil)
 	{
-		return obj;
+		assignNoAndUnregister = YES;
 	}
 	BOOL isGlobalObject = setObjectHasWeakRefs(obj);
-	// If we old ref exists, decrement its reference count.  This may also
-	// delete the weak reference control block.
-	if (oldRef != NULL)
-	{
-		weakRefRelease(oldRef);
-	}
-	// If we're storing nil, then just write a null pointer.
-	if (nil == obj)
-	{
-		*addr = obj;
-		return nil;
-	}
 	if (isGlobalObject)
 	{
 		// If this is a global object, it's never deallocated, so secretly make
 		// this a strong reference.
 		*addr = obj;
-		return obj;
+		return *addr;
 	}
-	if (nil != obj)
+	if(!assignNoAndUnregister)
 	{
+		// obj is not null, update it's registration
+		WeakRef *oldRef;
+		id old;
+		loadWeakPointer(addr, &old, &oldRef);
+		// If the old and new values are the same, then we don't need to do anything.
+		if (old == obj && obj != nil)
+		{
+			return *addr;
+		}
+		// release old ref
+		if (oldRef != NULL) {
+			weakRefRelease(oldRef);
+		}
+		// set new ref
 		*addr = (id)incrementWeakRefCount(obj);
 	}
-	return obj;
+	else {
+		*addr = nil;
+		WeakRef *oldRef;
+		id old;
+		if(loadWeakPointer(addr, &old, &oldRef)) {
+			weakRefRelease(oldRef);
+		}
+	}
+	return *addr;
 }
 
 OBJC_PUBLIC BOOL objc_delete_weak_refs(id obj)
@@ -914,16 +926,7 @@ OBJC_PUBLIC void objc_moveWeak(id *dest, id *src)
 
 OBJC_PUBLIC void objc_destroyWeak(id* obj)
 {
-	LOCK_FOR_SCOPE(&weakRefLock);
-	WeakRef *oldRef;
-	id old;
-	loadWeakPointer(obj, &old, &oldRef);
-	// If the old ref exists, decrement its reference count.  This may also
-	// delete the weak reference control block.
-	if (oldRef != NULL)
-	{
-		weakRefRelease(oldRef);
-	}
+	objc_storeWeak(obj, nil);
 }
 
 OBJC_PUBLIC id objc_initWeak(id *addr, id obj)
--- a/arc.m
+++ b/arc.m
@@ -199,8 +199,6 @@ extern BOOL FastARCAutorelease;
 
 static BOOL useARCAutoreleasePool;
 
-static int weakref_class;
-
 static const long refcount_shift = 1;
 /**
  * We use the top bit of the reference count to indicate whether an object has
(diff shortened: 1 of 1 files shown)


jordo:
Just another comment, as I'm not sure of the free() weak reference implications of this commit...  

But we were experiencing various crashes in objc_loadWeakRetained.  objc_loadWeakRetained seems to rely on the return value of loadWeakPointer and sometimes loadWeakPointer would return YES even though the isDeleted flag  in the weak pointer is true in weakRefGetObj.  I imagine loadWeakPointer needs to return NO in this scenario?

jordo:
So, i believe the current patch leaks weak references..  (not sure free() is called in some scenarios, and profiling seems to indicate so)...

But while debugging the weak ref table, and arc runtime spec, it brings up some more questions...    loadWeakPointer checks if there is "a strong (not deallocatable) object in a weak pointer"...  Is this really good?  Our app seems to have issues in objc_loadWeakRetained...  Is the fact that we're breaking contract with the arc contract potentially causing issues with the compiler?  The arc spec seems to say we should only return nil from that function when the passed in object has been deallocated or begun deallocation.  But we will return nil if the actual object is one of these (not deallocatable) objects inside a weak pointer.  I'm just wondering if the compiler is then making some optimizations and assumptions that could be breaking stuff...

jordo:
I hope noone minds but I'm kinda going to use this PR as a reference for where things are at for us in this weak ref table issue.  

I've finally been able to build on android with asan enabled and functioning correctly on arm64-v8a.  In my app, I can reliably reproduce a use-after-free error on master (even with the latest weakRefGetObj).   Considering this is cross-compiled and I don't have a nice way or environment to run tests on, I'm going to continue to gather as much information as I can.

After the latest patch, the use-after-free error just seems to be moved to a different function.  (see attached screenshot).  

Screen Shot 2019-08-19 at 9 34 33 PM






ERROR: AddressSanitizer: heap-use-after-free on address 0x00433b76bc10 at pc 0x0078c9bca4e8 bp 0x0078bd96dc30 sp 0x0078bd96dc28

jordo:
I'll symbolicate this.  It's a weird location, so I'm not sure whats going on, but attached for reference...

[asan_report.txt](https://github.com/gnustep/libobjc2/files/3518645/asan_report.txt)

jordo:
Alright, continuing on....

arc.m, in loadWeakPointer, classForObject takes an id, but I believe what's passed in is actually a weak_reference pointer (both structs have an isa field).  At any rate, after symbolication, the weak reference is already free'd somehow by something triggering weakRefRelease previously for this reference.

I'm still not sure how this is happening, but my only thought currently is there is a race condition where a weak_reference is getting it's 'isDeleted' flag set by objc_delete_weak_refs.  And the object address is getting re-used again for another weak reference before the original one is released and free'd in the runtime via weakRefRelease.

jordo:
Alrightly...  I think i've finally found this one.   (and a fix that isn't just leaking references)..

The just of it all, is I think objc_loadWeakRetained shouldn't actually ever decrement a weak reference count ( call weakRefRelease and potentially free a reference).  The arc spec actually doesn't say anything about freeing a reference in this function.  We're apparently just supposed to return nil in a scenario where an object is begun deallocation.

(I tested that this doesn't actually leak weak references by simply adding a weak_ref tracking counter in the calloc and free calls.)

I'll update this PR with a correct patch shortly

jordo:
In wrap up.   I'm really not sure how to write a test case for this...   The use-after-free issue was tricky to begin with, and the code that caused the crash is pretty simple:

`@property(nonatomic,readwrite,weak) CCNode* parent;`

And then calling this a bunch of times in a row:

`if(cell.parent) {`

It seems the compiler will just happily spit out a bunch of objc_loadWeakRetained which i don't think should ever free a weak ref under any condition.

jordo:
interesting... I didn't expect the objc_moveWeak test to fail, since that hasn't changed.  The test looks correct, but it doesn't seem to just be testing a objc_moveWeak as that wasn't changed.  

Investigating further, I replicated the test code here, and the line 'var = x;' isn't actually triggering a objc_moveWeak.  It calls objc_loadWeakRetained (at least on my compiler).

jordo:
OK.  so this test does completely different things on my compiler and android device.  Not only that, but it passes...
`assert(f1);

	assert([obj loadWeak] == nil);
	// Try to trigger an objc_moveWeak call
	{
		id o1 = [Test new];
		{
			id __weak x = o1;
			var = x;
		}
	}
	assert([obj loadWeak] == nil);`


obj_moveWeak is never called.  Here's what the compiler does on my device for this test:

[obj loadWeak]; implementation will call a objc_loadWeakRetained.

id __weak x = o1; will call a objc_initWeak

var = x; will call a objc_loadWeakRetained, then a objc_storeWeak  (NOTE NO objc_moveWeak)

} (first end curly brace) does a objc_destroyWeak

} (second end curly brace) does a objc_delete_weak_refs via objc_release_fast_np

[obj loadWeak]; implementation calls another objc_loadWeakRetained.

jordo:
For reference here's my compiler version:

Android (5657785 based on r353983d) clang version 9.0.4 (https://android.googlesource.com/toolchain/llvm-project 2a90cc9d899e6fe6dcf0da193244b25f75735c7b) (based on LLVM 9.0.4svn)

davidchisnall:
I think you are confused by reading the ARC spec.  It does not talk about weak counts, because that is an implementation issue.  The weak reference implementation works roughly like this:

- Each object has a retain count.  This is exposed as part of the ARC specification and is observable in the language in some ways.
- If an object has one or more weak references, then it has a `WeakRef` struct associated with it.
- The `WeakRef` struct is *also* refcounted, but this is not exposed as part of the ARC contract, it is an implementation detail.
- The refcount for the `WeakRef` struct is the number of weak pointers that exist to an object.

A `__weak id` is one of three things:

- `nil`.  This one is easy.
- A pointer to object such as an `NSString` or `Class` that is never deallocated and so we just store the strong pointer.  The runtime provides an API for opting classes into this behaviour if they are unsafe with weak references (or where they aren't useful, for example singleton instances that are never deallocated).
- A pointer to a `WeakRef` struct, disambiguated from a real pointer by its `isa` field.

When you do a load of a `__weak id` the runtime checks for the first two cases.  In either case, it returns the result.  In the final case, if the object is still live then it retains it and returns the result.  If the object has started deallocating, then its 'isDeleted' flag will be true.  The pointer that we're seeing is a weak pointer to a deallocated object.  In this case, we decrement the weak reference count and, if it reaches zero, deallocate the structure.  This is safe without atomics because we're holding a lock here.  That is a scalability bottleneck for highly concurrent code using a lot of weak pointers and we should probably move to a sharded lock model at some point if anyone cares. 

It sounds, from your report, as if the `WeakRef` is being prematurely deallocated (off-by-one error in the reference counting?).  I'm not sure where this would be though, because the only call to `weakRefRelease` that doesn't zero the memory afterwards is in `objc_destroyWeak`, where the contract is that the memory location won't be reused for a weak reference without calling `objc_initWeak` first.

jordo:
Hi David.  Thanks for the help and explanation!  That helps a bunch.  I've been trying to track down our weak reference instability for quite some time..

So, `weakRefRelease` in the current arc runtime implementation is a powerful func.  It is the decrementer for the `weakRef` struct in the implementation as you mentioned.  It can potentially unregister the ref/object and free() the `weakRef` struct (when the refcount == 0).

Right now, in the current arc.m implementation there are 3 arc functions that can potentially call `weakRefRelease`.  objc_storeWeak, objc_destroyWeak, and objc_loadWeakRetained.

The current implementation has a heap-use-after-free bug in `loadWeakPointer`.  If `loadWeakPointer` ever gets called with an `id* addr` that has already been unregistered and free'd in `weakRefRelease` we will have an invalid access in `classForObject` because that function reads the isa field.  Our app is single-threaded so there seems to exist some order of arc code generation that leads to code generation that leads to this bug.

At any rate, I'm still unsure why potentially unregistering the weak reference in `objc_loadWeakRetained` is necessary (is it?).  As it seems like it should always get cleaned up in a subsequent `objc_destroyWeak`...   My interpretation of the arc spec is that unregistering a weak ref should only be done in either `objc_destroyWeak` or `objc_storeWeak`.  

But yes, i agree that zeroing the memory after a `weakRefRelease` that deallocates the ref seems like it should guard against this.  It just doesn't seem to be the case for me right now.  (Also `weakRefRelease` in `objc_storeWeak` doesn't zero anything, so maybe related?).

Anyways, whats nice right now is I can consistently re-produce this issue in our app...  upon error asan does give me the three locations of of corruption.  (initial weakRef allocation point and thread.  deallocation point and thread.  And final use-after-free point and thread).

jordo:
Below are two screencaps further illustrating the scenario of the crash:

The first screencap shows a weakReference registered at 0x50e78 freeing the weakRef struct at 0xa0490 ( in a objc_loadWeakRetained ).  (weakRef struct refcount count must have equaled 0).


deallocation_by_weakref1

Shortly after that, the second screencap shows a different weakReference registered at 0xf2e78 pointing to the same weakRef struct at 0xa0490.  (Attemping again a objc_loadWeakRetained).  This second screencap crashes on line 850 with a heap-use-after-free.  (The weakRef struct at 0xa0490 was just deallocated).  

weak_ref_2

The zeroing of the memory on line 861 I guess won't guard access from a different registered weak reference on the same object?

jordo:
OK, so the last commit is a fix for objc_storeWeak and (hopefully) a correct implementation.

The spec for `id objc_storeWeak(id *object, id value)` says:
```
id objc_storeWeak(id *object, id value)

If value is a null pointer or the object to which it points has begun deallocation, object is assigned null and unregistered as a __weak object.
```

I found scenarios when incrementWeakRefCount was getting called from objc_storeWeak  and would find an existing weakRef with it's isDeleted flag is set to true.  (indicating the object has begun deallocation).  The current implementation of objc_storeWeak would still increment the refCount.  This fix should correctly handle those scenarios where objc_storeWeak is called with objects that have begun deallocation.

arc.m
```diff
@@ -767,39 +773,47 @@ static BOOL setObjectHasWeakRefs(id obj)
...
-	return obj;
+	else {
+		*addr = nil;
+		WeakRef *oldRef;
+		id old;
+		if(loadWeakPointer(addr, &old, &oldRef)) {
+			weakRefRelease(oldRef);
+		}
+	}
+	return *addr;
```
davidchisnall: The new logic is harder to follow, but I'm struggling to see what's actually different here.  Both versions are releasing the old weak ref if they find one (you're doing that on both code paths) and if the new value is not nil then incrementing its weak ref count and storing it in the variable.

arc.m
```diff
@@ -855,11 +869,6 @@ OBJC_PUBLIC id objc_loadWeakRetained(id* addr)
 	// will acquire the lock before attempting to deallocate)
 	if (obj == nil)
 	{
-		// If we've destroyed this weak ref, then make sure that we also deallocate the object.
```
davidchisnall: This change shouldn't matter for functional correctness, but you are removing an optimisation (i.e. dropping the refcount of the weak ref struct once we notice that it's no longer needed).

That said, looking at this code I believe that this is probably the real source of the bug that you're seeing.  We're unconditionally decrementing the refcount of we weakref structure, but we're then only removing our reference if it is the last weak reference.

arc.m
```diff
@@ -919,16 +928,7 @@ OBJC_PUBLIC void objc_moveWeak(id *dest, id *src)
 
 OBJC_PUBLIC void objc_destroyWeak(id* obj)
 {
-	LOCK_FOR_SCOPE(&weakRefLock);
```
davidchisnall: Again, removing optimisations to no benefit.

davidchisnall:
I now have a test case for the underlying issue and will try to commit a fix today.

arc.m
```diff
@@ -759,6 +760,11 @@ static BOOL setObjectHasWeakRefs(id obj)
 	else
 	{
 		assert(ref->object == obj);
+		// if the weakRef for obj isDeleted, it means obj has begun deallocation
+		if(ref->isDeleted) {
+			weakRefRelease(ref);
+			return NULL;
+		}
```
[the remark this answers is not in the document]
jordo: OK, I'll verify i hit this code-path, and post a stack trace in a bit.

arc.m
```diff
@@ -767,39 +773,47 @@ static BOOL setObjectHasWeakRefs(id obj)
...
+	BOOL assignNoAndUnregister = NO;
+	if(obj == nil)
 	{
-		return obj;
+		assignNoAndUnregister = YES;
 	}
 	BOOL isGlobalObject = setObjectHasWeakRefs(obj);
-	// If we old ref exists, decrement its reference count.  This may also
-	// delete the weak reference control block.
-	if (oldRef != NULL)
```
davidchisnall: Now that this is moved later, you are *not* releasing the `WeakRef` structure if assigning a global object to the weak object.  This may mask your bug, because we're now leaking `WeakRef`s on some code paths.

arc.m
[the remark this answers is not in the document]
jordo: so after a fresh clean & build, and with #124 instead of this patch applied i haven't hit this code path

jordo:
close this PR as I think most of it's handled in #124 ?

davidchisnall:
Closing this one, but it looks as if there's still a bug that #124 doesn't fix.

PR #124, opened 2019-08-21
Weak nil fix

davidchisnall:
Fix an issue where the WeakRef structure was having its reference count decremented incorrectly.

### DIFF
--- a/Test/CMakeLists.txt
+++ b/Test/CMakeLists.txt
@@ -37,6 +37,7 @@ set(TESTS
 	ResurrectInDealloc_arc.m
 	RuntimeTest.m
 	WeakBlock_arc.m
+	WeakRefLoad.m
 	WeakReferences_arc.m
 	WeakImportClass.m
 	ivar_arc.m
--- a/Test/WeakRefLoad.m
+++ b/Test/WeakRefLoad.m
@@ -1,0 +1,22 @@
+#include "Test.h"
+
+#define SIZE 5000
+
+int main(int argc, const char * argv[])
+{    
+	id t = [Test new];
+	id w1;
+	id w2;
+	objc_initWeak(&w1, t);
+	objc_initWeak(&w2, t);
+	[t release];
+	assert(objc_loadWeakRetained(&w1) == nil);
+	assert(objc_loadWeakRetained(&w2) == nil);
+	assert(w1 == nil);
+	assert(w2 == nil);
+	assert(objc_loadWeakRetained(&w1) == nil);
+	assert(objc_loadWeakRetained(&w2) == nil);
+	assert(w1 == nil);
+	assert(w2 == nil);
+	return 0;
+}
--- a/arc.m
+++ b/arc.m
@@ -855,7 +855,8 @@ OBJC_PUBLIC id objc_loadWeakRetained(id* addr)
 	// will acquire the lock before attempting to deallocate)
 	if (obj == nil)
 	{
-		// If we've destroyed this weak ref, then make sure that we also deallocate the object.
+		// If the object is destroyed, drop this reference to the WeakRef
+		// struct.
 		if (ref != NULL)
 		{
 			weakRefRelease(ref);
--- a/arc.m
+++ b/arc.m
@@ -856,8 +856,9 @@ OBJC_PUBLIC id objc_loadWeakRetained(id* addr)
 	if (obj == nil)
 	{
 		// If we've destroyed this weak ref, then make sure that we also deallocate the object.
-		if (weakRefRelease(ref))
+		if (ref != NULL)
 		{
+			weakRefRelease(ref);
 			*addr = nil;
 		}
 		return nil;


davidchisnall:
@jordo, I believe this should fix things for you.

jordo:
OK.  I will test this in about an hour

jordo:
Tested this (under asan again) and I still have heap corruption errors.   (Still triggered in objc_loadWeakRetained).

I set a breakpoint at the new NULL check to verify that codepath is taken in some scenarios (it is).
Screen Shot 2019-08-21 at 9 16 12 AM

However, there still exists memory corruption issues.   The same line of code later on will trigger a heap-use-after-free:

Screen Shot 2019-08-21 at 9 12 34 AM

davidchisnall:
The code that you have shown is quite dangerous, since you are reading from a __weak variable twice in the same scope (clang will warn about this, the fix is to check if `proxy` is `nil`, not `_proxy`).

Does your code work with valgrind?  The reduced test case passes now with valgrind, so this looks like a separate issue.  Can you see where the object that asan is reporting an error for was allocated, and also check whether the object containing the pointer to it was correctly zero-initialised?

jordo:
hmmm.  clang doesn't warn about that...  It seems safe to me as the code is synchronized, and on top of that our code is single-threaded anyways.    _proxy should only ever be nil or a valid weak reference.  The assignment from that weak ref to a strong ref should mean the weak ref should really always stay valid.  At any rate, changing the nil check to be on the strong ref rather than the weak ref didn't change anything.

jordo:
I'm not familiar with running valgrind on android.  It'll probably take me a while to try and get that working.  Considering how annoying it is to get asan running correctly, i imagine it may take some time.  I'll see what I can do.

With regards to asan, and the heap use after free...  Asan does track the allocation point, free point, and use after free point (stacks and threads).  I posted that in https://github.com/gnustep/libobjc2/pull/123 but didn't symbolicate it.  At a quick glance the allocation is objc_initWeak.  free is objc_loadWeakRetained, and use after free is also in objc_loadWeakRetained.

davidchisnall:
Can you trace the refcount manipulations on the `WeakRef` (or, at least, stick in a `printf` to log the argument to `objc_loadWeakedRetained`)?  We should be freeing the `WeakRef` in a call to `objc_loadWeakRetained`, but only if the last `__weak` pointer is destroyed / set to `nil`, so I'd like to understand how the second call to `objc_loadWeakRetained` is reaching the same `WeakRef` as the first.  Do you have two `__weak` pointers to the same object, or is it two reads of the same variable? I have a test case for both of these that is now passing with this branch, so I want to understand what is different in your usage that is triggering the bug.

jordo:
OK, I will add some more manual tracing...  It is worthwhile to have some profiling/debugging flag in the arc runtime that can track the total weakRef counts?  We could then compile/enable it for a test to try and ensure weakRefs are not leaked.  

I have a rudimentary one for testing right now, that just tracks a count when we alloc a weakRef struct, and dealloc one.

jordo:
Hmmm, looks like valgrind has been deprecated from AOSP in favor of asan.  And to get asan running I had to build a eng flavor build of AOSP:  

Screen Shot 2019-08-21 at 10 48 54 AM

jordo:
So I'm fairly certain, this patch fixes at least one of our issues...  The optimization makes sense, looks correct now,  and the corruption that I had been able to easily replicate is gone with this patch.

Our app will still trigger a random heap corruption after a while of use, but i can't replicate it reliably like I could before this patch.

jordo:
@davidchisnall   arc.m line 679:

```*ref = (WeakRef*)oldObj;```

is this cast correct?  oldObj should be a pointer to id, not to WeakRef, no?

**Comment by jordo** [comment]
asan heap corruption spots (not sure how this one is possible):

alloc:  
 - allocate_class
dealloc:
 - free_object

use-after-free:
 - objc_loadWeakRetained

**Comment by jordo** [comment]
So, some more information on the memory corruption...  The allocation and free seem to happen on a regular (__strong) object, but the memory corruption occurs on a weak ref.  This is the problematic code as mentioned in other PR:
```
176: - (CCProxy *)proxy
177:    {
178: 	@synchronized(self){
179: 		__strong CCProxy *proxy = _proxy;
180: 
181: 		if (proxy == nil){
182: 			proxy = [[CCProxy alloc] initWithTarget:self];
183:			_proxy = proxy;
184:		}
185:
186:		return(proxy);
187:	}
189: }
```
When the corruption is detected, here are the allocation, deallocation, and error points:

```
allocation stack:
-----------------------------------
allocate_class (gc_none.c:19)
class_createInstance (runtime.c:344)
NSAllocateObject (NSObject.m:775)
_c_NSObject__allocWithZone_ (NSObject.m:1161)
_c_NSObject__alloc (NSObject.m:1122)
_i_CCSpriteFrame__proxy (CCSpriteFrame.m:182)  // the strong assignment in the above code
....
```
```
free stack:
-----------------------------------
free_object (gc_none.c:29)
object_dispose (runtime.c:362)
NSDeallocateObject (NSObject.m:859)
_i_NSObject__dealloc (NSObject.m:1357)
_i_CCProxy__dealloc (CCTexture.m:167)  // seems like a regular object deallocation
```
```
use after free stack:
-----------------------------------
classForObject (class.h:423) // crash line
loadWeakPointer
objc_loadWeakRetained
_i_CCSpriteFrame__proxy (CCSpriteFrame.m:179) // somehow same memory address is used here for a weak ref load
```

Below is the memory quarantine regions reported by asan when corruption is detected.  What's interesting is there seems to be a 3-byte region that's free'd ([fd]), and we try to read the second byte?:
```
 SUMMARY: AddressSanitizer: heap-use-after-free 
 Shadow bytes around the buggy address:
   0x001867396c40: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
   0x001867396c50: fa fa fa fa fa fa fa fa fa fa fd fd fd fa fa fa
   0x001867396c60: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
   0x001867396c70: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
   0x001867396c80: fa fa fd fd fd fd fd fd fa fa fa fa fa fa fa fa
 =>0x001867396c90: fa fa fa fa fa fa fa fa fa fa fd[fd]fd fa fa fa
   0x001867396ca0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
   0x001867396cb0: fa fa fa fa fa fa fa fa fa fa 00 00 00 00 00 00
   0x001867396cc0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
   0x001867396cd0: fa fa fd fd fd fd fd fd fa fa 00 00 00 fa fa fa
   0x001867396ce0: fa fa 00 00 00 00 00 fa fa fa fa fa fa fa fa fa
 Shadow byte legend (one shadow byte represents 8 application bytes):
   Addressable:           00
   Partially addressable: 01 02 03 04 05 06 07 
   Heap left redzone:       fa
   Freed heap region:       fd
   Stack left redzone:      f1
   Stack mid redzone:       f2
   Stack right redzone:     f3
   Stack after return:      f5
   Stack use after scope:   f8
   Global redzone:          f9
   Global init order:       f6
   Poisoned by user:        f7
   Container overflow:      fc
   Array cookie:            ac
   Intra object redzone:    bb
   ASan internal:           fe
   Left alloca redzone:     ca
   Right alloca redzone:    cb
   Shadow gap:              cc
 ==14752==ABORTING
```

I will say, since this patch the above error is pretty rare now...  I have to try pretty hard to get it to happen, and when it does happen, it seems to be purely random.  But it is always the same spot in our application code.

jordo:
@davidchisnall Continuing down this rabbit hole...  It seems like the arguments to calloc in gc_none.c are backwards?  Is there any potential to mess up alignment somehow here?

davidchisnall:
In #123, there is a bug that doesn't decrement the weak count for an object when assigning a non-refcounted object to a weak pointer.  I wonder if this is hiding the issue for you?

I don't believe order of arguments for calloc matters.  Every implementation that I've seen multiplies the two together, checks for overflow, and forwards to malloc.  

Your call stack at the crash is a little bit odd.  It's crashing in `loadWeakPointer` and the only memory that this accesses (other than the stack) is the pointer that it's passed (should be the `__weak` ivar of your object) and, if that is a `WeakRef` then a field of the `WeakRef`.  It should never read the target object (unless the object is not refcounted, in which case it would read its `isa` field).  It looks as if trying to read the `isa` field of whatever is pointed to by your field is causing the crash.  There are two explanations I can see for your behaviour:

- Something has accidentally stored a pointer to the object, not to the `WeakRef` in your ivar.  When the object was deallocated, this became dangling and everything broke.
- The `WeakRef` was prematurely deallocated (either the `WeakRef` is deallocated when it should be but the pointer to it is not zeroed, or the `WeakRef` is deallocated too early).

The `WeakRef` structure is 24 bytes (an object with a single field is also 24 bytes - is that the case for `CCProxy`?), so these may end up sharing the same address if one is deallocated before the other (e.g. if there's a `WeakRef`, it's deallocated with a dangling pointer, there's a `CCProxy` allocated at the same address, the `CCProxy` is deallocated, and this confuses ASan.  This is why I prefer valgrind for debugging this kind of thing: ASan is a lot faster than valgrind, but also somewhat less accurate.

You could add an ivar to `CCProxy` to make it a different size to help disambiguate these two cases.  

There are only three calls to `weakRefRelease` and two  of them are immediately followed by a store to the memory location (with a different value), the other is in `objc_destroyWeak`, which has the invariant that the caller must not reuse the memory for a weak pointer without first calling `objc_initWeak` on it (you can try adding a `*obj = nil` there and see if that helps - if so, it may be a compiler bug.

Anything you can do to get a test case that fails reliably (ideally, send a PR to add something to the libobjc2 test suite) would be very valuable.

PR #127, opened 2019-10-03
Fixed crash when using self in imp_implementationWithBlock on armv7a

### LINKED ISSUE #126: Crash when using self in imp_implementationWithBlock on armeabi-v7a
On armeabi-v7a, when you try and use self in a imp_implementationWithBlock block, the app crashes.

It works correctly for arm64-v8a, x86 and x86_64.

Attached is a patch to BlockImpTest.m that can be used to reproduce the issue.
[libobjc-block-imp-crash.patch.zip](https://github.com/gnustep/libobjc2/files/3681196/libobjc-block-imp-crash.patch.zip)

rupertdaniel:
These changes fix #126

### DIFF
--- a/Test/BlockImpTest.m
+++ b/Test/BlockImpTest.m
@@ -26,6 +26,7 @@ int main(void)
 {
 	__block int b = 0;
 	void* blk = ^(id self, int a) {
+		assert([self class] == [Foo class]);
 		b += a; 
 		return b; };
 	blk = Block_copy(blk);
@@ -45,6 +46,7 @@ int main(void)
 	assert(imp_getBlock(imp) != (blk));
 
 	blk = ^(id self) {
+		assert([self class] == [Foo class]);
 		struct big b = {1, 2, 3, 4, 5};
 		return b;
 	};
--- a/block_trampolines.S
+++ b/block_trampolines.S
@@ -126,7 +126,7 @@
 .thumb
 .macro trampoline arg0, arg1
 	sub r12, pc, #4095
-	mov \arg0, \arg1            // Move self over _cmd
+	mov \arg1, \arg0            // Move self over _cmd
 	ldr \arg0, [r12, #-5]       // Load the block pointer over self
 	ldr r12, [r12, #-1]         // Jump to the block function
 	bx  r12
@@ -134,7 +134,7 @@
 #	else
 .macro trampoline arg0, arg1
 	sub r12, pc, #4096
-	mov \arg0, \arg1            // Move self over _cmd
+	mov \arg1, \arg0            // Move self over _cmd
 	ldr \arg0, [r12, #-8]       // Load the block pointer over self
 	ldr pc, [r12, #-4]          // Jump to the block function
 .endm


Test/BlockImpTest.m
```diff
@@ -26,6 +26,7 @@ int main(void)
 {
 	__block int b = 0;
 	void* blk = ^(id self, int a) {
+		assert([self class] == [Foo class]);
```
davidchisnall: Please can you put the result of `[foo class]` in a local variable before you create the block and then just compare `self` against that global (if `self` is incorrect then we will end up with a crash in `objc_msgSend` for the `[self class]` message, rather than a helpful assert that `self` is wrong.

Test/BlockImpTest.m
```diff
@@ -45,6 +46,7 @@ int main(void)
 	assert(imp_getBlock(imp) != (blk));
 
 	blk = ^(id self) {
+		assert([self class] == [Foo class]);
```
davidchisnall: Same change here.

Test/BlockImpTest.m
rupertdaniel: I just tried doing this, and `assert(self == globalFooClass)` always fails. Am I doing the comparison incorrectly?

davidchisnall: Should work if `globalFooClass` is initialised before the block is created.  What are the two values?  I would simply move line 36 earlier and use that, because that's the value that we're explicitly passing to the first `objc_msgSend` call.

rupertdaniel: They both have the same memory address when I printf them e.g.
```
fooClass: 8081d35c
self: 8081d35c
```
but the `assert(self == fooClass)` doesn't like it.

davidchisnall: Can you post the complete diff?

rupertdaniel:
```
diff --git a/Test/BlockImpTest.m b/Test/BlockImpTest.m
index ce9931a..7fea7ad 100644
--- a/Test/BlockImpTest.m
+++ b/Test/BlockImpTest.m
@@ -24,8 +24,11 @@ +(struct big)sret;
 
 int main(void)
 {
+	Class cls = objc_getClass("Foo");
+	
 	__block int b = 0;
 	void* blk = ^(id self, int a) {
+		assert(self == cls);
 		b += a; 
 		return b; };
 	blk = Block_copy(blk);
@@ -33,7 +36,6 @@ int main(void)
 	char *type = block_copyIMPTypeEncoding_np(blk);
 	assert(NULL != type);
 	class_addMethod((objc_getMetaClass("Foo")), @selector(count:), imp, type);
-	Class cls = objc_getClass("Foo");
 	assert(2 == ((int(*)(id,SEL,int))imp)(cls, @selector(count:), 2));
 	free(type);
 	assert(4 == [Foo count: 2]);
@@ -45,6 +47,7 @@ int main(void)
 	assert(imp_getBlock(imp) != (blk));
 
 	blk = ^(id self) {
+		assert(self == cls);
 		struct big b = {1, 2, 3, 4, 5};
 		return b;
 	};
```

davidchisnall: Does this fail in the first or second assert?  If you print both values in the block before the assert, what does it show?

rupertdaniel: It fails on the first assert.

If I printf before the first assert:
```
self: 7fcd135c cls: 0
```
So I guess the `0` is the problem.

rupertdaniel: So if I declare

Test/BlockImpTest.m
```diff
@@ -24,17 +24,17 @@ +(struct big)sret;
 
 int main(void)
 {
+	__block Class cls = objc_getClass("Foo");
```
davidchisnall: This shouldn't need to be `__block`, it should be fine to capture by value.  It it isn't, then we should figure out why not.

Test/BlockImpTest.m
davidchisnall: That sounds as if there may be a bug in the block copy, but for me clang emits a copy helper with a call to `_Block_object_assign` with and without the `__block` qualifier.

rupertdaniel: What is the best way to check this on my side?

davidchisnall: Can you put a breakpoint on `_Block_object_assign` and check that, in the `imp_implementationWithBlock` call it's being called for `cls`?  It's also possible that the block trampoline is not actually loading the block into the first argument register correctly.  Do you see the same failure with the ARM and Thumb versions?  The load to `\arg0` is supposed to be from exactly 4 bytes before the block pointer that's called.  It's difficult to check this directly, but you can create a block structure directly using the struct from the `blocks_private.h` header and give it a function pointer that checks the address of the block variable.  We should probably have a test that does that anyway...

davidchisnall:
I'm going to merge this to get the fix, but it sounds as if the test case was picking up something else and I'd like to understand what is still broken.

Issue #108, opened 2019-05-21
ObjCXXEHInterop tests fail on Android


triplef:
Running the ObjCXXEHInterop* tests on Android ARMv7 fails with the following messages (libobjc2 was compiled with `DEBUG_EXCEPTIONS=1`). The same tests succeed on ARM64.

```
RUNNING ObjCXXEHInterop...
Poking from minRepM
Raising MyException
Throwing 0xea817014
Caught - re-raising
Throwing 0xea817014
New personality function called 0xea847080
Class: GNUCOBJC
LSDA: 0xac8dcd10
Search phase...
Filter: 1
Class name: Test
0xac8df1d0 type: 1
found handler for Test
handler: 4
Found handler! 4
Aborted 
FAILED: ObjCXXEHInterop

RUNNING ObjCXXEHInterop_legacy...
Poking from minRepM
Raising MyException
Throwing 0xee317014
Caught - re-raising
Throwing 0xee317014
New personality function called 0xee355080
Class: GNUCOBJC
LSDA: 0xb2111d10
Search phase...
Filter: 1
Class name: Test
0xee3420a0 type: 1
found handler for Test
handler: 4
Found handler! 4
Aborted 
FAILED: ObjCXXEHInterop_legacy

RUNNING ObjCXXEHInterop_legacy_optimised...
Poking from minRepM
Raising MyException
Throwing 0xf2397014
Caught - re-raising
Throwing 0xf2397014
New personality function called 0xf23d5080
Class: GNUCOBJC
LSDA: 0xb9c71d2c
Search phase...
Filter: 1
Class name: Test
0xf23c20a0 type: 1
found handler for Test
handler: 4
Found handler! 4
Aborted 
FAILED: ObjCXXEHInterop_legacy_optimised

RUNNING ObjCXXEHInterop_optimised...
Poking from minRepM
Raising MyException
Throwing 0xef817014
Caught - re-raising
Throwing 0xef817014
New personality function called 0xef847080
Class: GNUCOBJC
LSDA: 0xb2f5ed18
Search phase...
Filter: 1
Class name: Test
0xb2f611d0 type: 1
found handler for Test
handler: 4
Found handler! 4
Aborted 
FAILED: ObjCXXEHInterop_optimised
```

As I’m not really familiar with Objective C/C++ exception handling mechanisms I’d appreciate any pointers to track this down further.

[...]

PR #129, opened 2019-11-12
Added referenceCount to __cxa_exception struct and updated unwind alignment

triplef:
Updated `__cxa_exception` to add `referenceCount` fields to match lates revisions from libcxxabi and libcxxrt:
https://github.com/llvm/llvm-project/blob/master/libcxxabi/src/cxa_exception.h#L30-L67
https://github.com/pathscale/libcxxrt/blob/master/src/cxxabi.h#L77-L165

Updated `_Unwind_Exception` structs to match latest revisions from libunwind:

- ARM version is using `__attribute__((__aligned__(8)))`:  
https://github.com/llvm/llvm-project/blob/master/libunwind/include/unwind.h#L107

- Itanium version is using `__attribute__((__aligned__))`, `uintptr_t` for private fields, and an added `reserved` field to be explicit about alignment:
https://github.com/llvm/llvm-project/blob/master/libunwind/include/unwind.h#L123-L143

The last two changes are also mirrored in this PR for libcxxrt:
https://github.com/libcxxrt/libcxxrt/pull/1

Fixes ObjCXXEHInterop tests on Android x86 and x86_64, as reported in #108. However, it does not yet fix the same test on Android ARMv7.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,11 +1,27 @@
 freebsd_instance:
   image: freebsd-12-0-release-amd64
 
-task:
+libcxxrt_freebsd_task:
   install_script: pkg install -y cmake ninja llvm80
-  script: |
+  build_script: |
       mkdir Build
       cd Build
       cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang80 -DCMAKE_CXX_COMPILER=clang++80
       ninja
-      ctest -j4
+  test_script: cd Build && ctest -j4
+
+libcxxrt_master_task:
+  install_script: pkg install -y git cmake ninja llvm80
+  install_libcxxrt_script: |
+      git clone https://github.com/libcxxrt/libcxxrt.git
+      mkdir -p libcxxrt/Build
+      cd libcxxrt/Build
+      cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang80 -DCMAKE_CXX_COMPILER=clang++80
+      ninja
+      cp lib/libcxxrt.so /usr/local/lib
+  build_script: |
+      mkdir Build
+      cd Build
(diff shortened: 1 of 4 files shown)


triplef:
Looks like the tests on Cirrus CI are failing, because it would need to be tested with the updated libcxxrt containing this PR: https://github.com/pathscale/libcxxrt/pull/54.

The tests on Azure seem to be using the C++ standard library instead.

davidchisnall:
Please can you send the PR to libcxxrt/libcxxrt?  We moved the canonical upstream for libcxxrt there a short time ago.

davidchisnall:
I am quite nervous about committing this, because the fact that it breaks in CI means that it is probably an ABI-breaking change.  I'd like to get @emaste's opinion on the correct way to handle this.  We probably want to add some conditional compilation both here and in libcxxrt to allow the old ABI to remain supported, and add a check in CMake here for what the C++ runtime library does.

triplef:
Thanks @davidchisnall, I’d appreciate if you can take a look at this. I can’t say I fully understand in how far libobjc2, libcxxrt, and libcxxabi, and libunwind need to be aligned, but these changes did fix ObjC++ interop for us on x86 and x86_64. The main problem was that for us `sizeof(__cxa_exception)` didn’t match between libobjc2 and libcxxrt.

I’ve resubmitted the libcxxrt PR to the new repo and updated the links above.

davidchisnall:
I don't think working around ABI incompatibilities by installing a different version is the correct fix.  We should still be doing CI on FreeBSD with the built versions, we need to make sure that we expose an option to select the ABI and, ideally, a test that checks for the correct one.

triplef:
@davidchisnall I set up the FreeBSD CI now to test against both libcxxrt from FreeBSD and the latest libcxxrt master (which I think makes sense regardless of this PR – let me know if I should submit this as a separate PR). As outlined in https://github.com/libcxxrt/libcxxrt/pull/1#issuecomment-555224884, this would currently only pass against libcxxrt together with https://github.com/libcxxrt/libcxxrt/pull/1.

triplef:
This was obsoleted by https://github.com/gnustep/libobjc2/commit/242442b3aa1d0e438a6893ba5d8b472c9634d293 and #132.

PR #130, opened 2019-11-14
Match _Unwind_Reason_Code enum between ARM and Itanium.

triplef:
Unwind ARM was using `_URC_OK` and Itanium `_URC_NO_REASON`, which made it difficult to write code that worked with both versions. This adds the missing one to each, which matches what is done in libunwind:
https://github.com/llvm/llvm-project/blob/master/libunwind/include/unwind.h#L33-L34

Also fixes some mismatched in indentation.


triplef:
Looks like the `ManyManySelectors_optimised` test was hanging on the "Windows-2016 Debug-32" CI. Any idea why that might have been? Doesn’t seem to me like this could be related to this change.

PR #134, opened 2019-11-26
generate a pkg-config file

ngrewe:
This is quite easy to do and allows build systems (even autoconf, if you want to call it a build system…) to discover a libobjc dependency more conveniently.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -422,6 +422,21 @@ if (UNIX)
 endif ()
 include (CPack)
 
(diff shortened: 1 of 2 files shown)


davidchisnall:
I'm happy to merge this, but I'm not entirely sure that it's useful.  The pkgconfig format doesn't provide a way of describing Objective-C flags and we should probably add `-fblocks -fobjc-runtime=gnustep-{version}` there.

PR #135, opened 2019-11-29
Fix #ifdef for class property test

ngrewe:
Minor cleanup: The `#ifdef` was testing the wrong macro.

### DIFF
--- a/Test/category_properties.m
+++ b/Test/category_properties.m
@@ -31,9 +31,9 @@ int main(int argc, char** argv)
 	prop = class_getProperty(test, "val2");
 	assert(prop);
 	assert(strcmp("Ti,R,D", property_getAttributes(prop)) == 0);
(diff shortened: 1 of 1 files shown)


davidchisnall:
This isn't duplicated code, this code is checking the class property.  It's also not an ineffective ifdef: it ensures that this part of the test runs only with the v2 ABI.

ngrewe:
Ah, sorry, I‘m missed the `object_getClass` call — the ifdef is still broken, though (extra `D` in front of the variable name, redeclaration of `prop` inside the Block).

ngrewe:
Any thoughts, David? I still think there is an issue with this test…

davidchisnall:
I think the only thing that's a problem here is the redefinition of prop.

ngrewe:
Oops, totally screwed up the rebase this morning. Now it should reflect what's intended again.

PR #149, opened 2020-02-20
Fix detection of ARC-incompatible memory management methods

rfm:
check for release and autorelease as well as for retain


davidchisnall:
Looks fine, please can you fix the commit message (force push to the branch to overwrite it) and add a test (add a release method that sets a flag to a class and then call `objc_release` on an instance of it and check that it is called).

Issue #146, opened 2020-02-11
Crash when throwing exceptions in ObjC++ due to __cxa_exception struct mismatch


triplef:
Throwing exceptions in Objective C++ code currently crashes with some setups. #138 adds tests that expose (some of?) these issues.

According to @davidchisnall:

> I've had a look and found a few minor bugs that are easy to fix. Unfortunately, when I fix them I find some more interesting issues that arise from the way that we're now creating the exception by hijacking the C++ runtime library.
> I think that the solution that we're going to have to end up with is using that path only to figure our which structure we have and then casting to the correct variant structure and creating it ourselves.

I have created [cxa_exception_variants.cpp](https://github.com/gnustep/libobjc2/files/4204619/cxa_exception_variants.cpp.zip) containing the four different possible layouts of the __cxa_exception struct that are found in libobjc2, libcxxrt, and libunwind:
* There are two different versions of the __cxa_exception struct: without and with the added "referenceCount" field.
* However the struct also references _Unwind_Exception, which itself comes in two different variants: without and with an "\_\_aligned\_\_" attribute (and an added "reserved" field on x86).

This leaves us with a total of four different variants of __cxa_exception:

- __cxa_exception_v1: without referenceCount field, using unaligned _Unwind_Exception
- __cxa_exception_v2: without referenceCount field, using aligned _Unwind_Exception
- __cxa_exception_v3: with referenceCount field, using unaligned _Unwind_Exception
- __cxa_exception_v4: with referenceCount field, using aligned _Unwind_Exception

Depending on the architecture some of these will have the same size. Following are the sizes for the different architectures as printed by the attached tool.

**x86**
```
_Unwind_Exception_v1	size: 20, alignment: 4
_Unwind_Exception_v2	size: 32, alignment: 16

__cxa_exception_v1	size: 64, alignment: 4
__cxa_exception_v2	size: 80, alignment: 16
__cxa_exception_v3	size: 68, alignment: 4
__cxa_exception_v4	size: 80, alignment: 16
```

**x86_64**
```
_Unwind_Exception_v1	size: 32, alignment: 8
_Unwind_Exception_v2	size: 32, alignment: 16

__cxa_exception_v1	size: 112, alignment: 8
__cxa_exception_v2	size: 112, alignment: 16
__cxa_exception_v3	size: 120, alignment: 8
__cxa_exception_v4	size: 128, alignment: 16
```

**ARM32**
```
[...]

PR #156, opened 2020-04-26
Initial run-time detection of C++ EH ABI variant.

davidchisnall:
On Itanium C++ ABI platforms, there are a lot of different variations
that have subtly different structure layouts.  This commit adds a
run-time test that throws a C++ exception through a function that has
its personality function replaced with a wrapper so that we can inspect
a C++ exception in controlled conditions.

Initially, this just detects the exception type value used for C++
exceptions.


davidchisnall:
@ngrewe, this should auto-detect whether to use `"CLNGC++\0"` or `"GNUCC++\0"` as the exception type for C++.  I intend to use this logic to pick between exception structure layouts from the list in #146.

davidchisnall:
@triplef, this now does dynamic detection of the C++ exception structure layout and so should be more robust in the presence of ABI changes in the future.

@ngrewe, hopefully the dynamic detection used here can simplify your work to support libc++abi.

PR #157, opened 2020-04-30
Documentation: Add INSTALL instructions 

Letterus:
to reflect that git submodules need to be checked out as well.


davidchisnall:
Thank you.  Ideally, we'd have a CMake check that would do the submodule init thing if the third_party directory didn't contain the right things, but I haven't had a chance to write one.

PR #178, opened 2020-08-26
Don't use CXXFLAGS when compiling eh_trampoline.cc

### LINKED ISSUE #177: Can't link certain programs with libobjc: undefined reference: eh_trampoline()
Hi!
After updating the FreeBSD port of libobjc2 to the latest version, specific programs such as gnustep-back, can no longer link with libobjc2.  I get the error 
```
/usr/local/bin/ld.bfd: /usr/local/lib/libobjc.so: undefined reference to `eh_trampoline()'
cc: error: linker command failed with exit code 1 (use -v to see invocation)
```
see http://package18.nyi.freebsd.org/data/121amd64-default-PR244494/2020-08-25_21h45m10s/logs/errors/gnustep-back-0.27.0.log for more details.

I'm unsure about the cause of this. eh_trampoline() is declared as hidden, maybe it shouldn't be? I can't rule out it's specific compiler flags in either the build of libobjc2 or gnustep-back that's the problem, but I'm unfamiliar with gnustep in general and would muc appreciate some guidance.
Thank you!

davidchisnall:
The way that we were doing this didn't handle multiple flags and we
actually don't want the user to override these flags because that file
needs to be compiled in a very specific way.

Fixes #177

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,7 +317,7 @@ if (ENABLE_OBJCXX)
 			endif()
 		endif ()
 		add_custom_command(OUTPUT eh_trampoline.s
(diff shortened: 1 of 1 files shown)


zeising (approved):
This solves my issue as described in #177 .
Thank you very much for your help, and for the quick resolution!

PR #181, opened 2020-09-03
Add pthreads to link libraries, not link flags

### LINKED ISSUE #180: Linking issue with --as-needed
Running tests with 2.1 on Gentoo Linux (libobjc will depend on C++ standard library), linked libobjc.so misses pthread:

```
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libobjc.so.4.6: undefined reference to `pthread_mutexattr_destroy'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libobjc.so.4.6: undefined reference to `pthread_key_create'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libobjc.so.4.6: undefined reference to `pthread_getspecific'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libobjc.so.4.6: undefined reference to `pthread_mutexattr_settype'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libobjc.so.4.6: undefined reference to `pthread_setspecific'
/usr/lib/gcc/x86_64-pc-linux-gnu/10.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libobjc.so.4.6: undefined reference to `pthread_mutexattr_init'
clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
```

This comes from CMakeLists.txt which adds pthread link libraries to link flags option:
https://github.com/gnustep/libobjc2/blob/master/CMakeLists.txt#L347

This change works fine locally for linking (some tests fail but this is another issue):
```
diff -Naur libobjc2-2.1.orig/CMakeLists.txt libobjc2-2.1/CMakeLists.txt
--- libobjc2-2.1.orig/CMakeLists.txt	2020-08-23 17:10:20.000000000 +0200
+++ libobjc2-2.1/CMakeLists.txt	2020-09-03 10:51:38.716078009 +0200
@@ -344,7 +344,7 @@
 # threading implementation (we do for everything except thread-local storage)
 set(CMAKE_THREAD_PREFER_PTHREAD)
 include(FindThreads)
-set(objc_LINK_FLAGS "${objc_LINK_FLAGS} ${CMAKE_THREAD_LIBS_INIT}")
+target_link_libraries(objc Threads::Threads)
```

With this, -lpthread gets added to correct variable:

>   LINK_FLAGS = -Wl,-O1 -Wl,--as-needed
>   LINK_LIBRARIES = -lpthread  -lstdc++  -lm

voyageur:
This fixes linking errors with --as-needed like:
ld: libobjc.so.4.6: undefined reference to `pthread_mutexattr_destroy'

Fixes #180

The cmake target Threads::Threads support was apparently added in cmake 3.1 which is the minimum required version set in CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -353,7 +353,7 @@ endif()
 # threading implementation (we do for everything except thread-local storage)
 set(CMAKE_THREAD_PREFER_PTHREAD)
 include(FindThreads)
-set(objc_LINK_FLAGS "${objc_LINK_FLAGS} ${CMAKE_THREAD_LIBS_INIT}")
+target_link_libraries(objc Threads::Threads)
 
 
 


davidchisnall:
This seems to have broken the FreeBSD build.  It looks as if it's not adding a required `-lm`.

voyageur:
Hmm it came with the pthreads flags here "-lpthread  -lstdc++  -lm", let me push a verbose build to see the linking command

Worst case, a less pretty fix can be to use both link flags and link libraries

voyageur:
OK so forcing both lines does not do the trick, nor adding hardcoded "-lm" to link libs (also could not enable verbose logs).

Reading error again, it may be that now with clean flags, this uncovers test binaries need math lib (with proper detection as hardcoded one break windows builds)

voyageur:
Note the missing "__throw_length_error()" also shows missing -lstdc++, so not only -lm

davidchisnall:
On FreeBSD the default is libc++, but your point still stands. It looks as if the test is being linked with clang, not clang++. We must be setting the linker language to C for the tests, even when they contain C++. I’m not sure why we’re missing libm, because we should only need it indirectly via libc++.

I’m also a bit confused about where the std::vector use comes from. We shouldn’t be depending on this in the library and I don’t think the tests use it. I’ll try to take a look over the weekend.

davidchisnall:
#182 should fix the issue with the `std::vector` symbols, but I don't quite understand why your change causes the problem with `ceilf`.  It looks as if that function is always undefined in `libobjc.so` and the tests do not link libm, so I don't know why this works before your change.  We probably do need to explicitly link libm on platforms where it exists: `std::ceil` is used by the `robin_hash` implementation and, at least with libc++, that's a thin wrapper around `ceilf`.

voyageur:
Hopefully that last part will be fine once the libm checks parts in libc++abi are merged, I did not think a small change like that would open such a can of worms :)

davidchisnall:
The libc++abi changes are merged and we're also now testing a lot more configurations in CI.  Please can you rebase this branch?  Thank you!

voyageur:
Sure, let's see how it goes!

davidchisnall (approved):

PR #183, opened 2020-09-15
Use CMAKE_AS_COMPILER to compile eh_trampoline.s.

triplef:
Fixes the following errors when building for Android with `AS` set to something like `bin/arm-linux-androideabi-as`:

```
[  9%] Building ASM object CMakeFiles/objc.dir/eh_trampoline.s.o
clang: warning: argument unused during compilation: '-D CXA_ALLOCATE_EXCEPTION_SPECIFIER=noexcept' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D GC_DEBUG' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D GNUSTEP' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D NO_LEGACY' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D NO_OBJCXX' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D TYPE_DEPENDENT_DISPATCH' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D __OBJC_RUNTIME_INTERNAL__=1' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D objc_EXPORTS' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D ANDROID' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-fdata-sections' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-ffunction-sections' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-funwind-tables' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-fstack-protector-strong' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D _FORTIFY_SOURCE=2' [-Wunused-command-line-argument]
clang: warning: argument unused during compilation: '-D NDEBUG' [-Wunused-command-line-argument]
eh_trampoline.s:1:18: error: expected string in directive
        .section        __TEXT,__text,regular,pure_instructions
                               ^
eh_trampoline.s:2:2: error: unknown directive
        .macosx_version_min 10, 15
        ^
eh_trampoline.s:3:18: error: unable to emit symbol attribute in directive
        .private_extern __Z13eh_trampolinev ## -- Begin function _Z13eh_trampolinev
                        ^
...
```


davidchisnall:
The change doesn't seem related to the error.  The line you've changed is to do with generating the .s file from the .cpp file.  The errors appear to be related to generating the .o file from the .s file.

triplef:
Yeah you’re right... I didn’t notice that with this change, generating the .s file fails and leaves the file empty, and that empty assembly file compiles fine. ;)

Any idea what might be causing these errors? The full build command is as follows:
```
[  1%] Building ASM object CMakeFiles/objc.dir/eh_trampoline.s.o
~/Library/Android/sdk/ndk/21.3.6528147/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang -target armv7-none-linux-androideabi21 -gcc-toolchain ~/Library/Android/sdk/ndk/21.3.6528147/toolchains/llvm/prebuilt/darwin-x86_64 --sysroot=~/Library/Android/sdk/ndk/21.3.6528147/toolchains/llvm/prebuilt/darwin-x86_64/sysroot -DCXA_ALLOCATE_EXCEPTION_SPECIFIER=noexcept -DGC_DEBUG -DGNUSTEP -DNO_LEGACY -DNO_OBJCXX -DTYPE_DEPENDENT_DISPATCH -D__OBJC_RUNTIME_INTERNAL__=1 -Dobjc_EXPORTS  -g -DANDROID -fdata-sections -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -D_FORTIFY_SOURCE=2 -march=armv7-a -mthumb -Wformat -Werror=format-security  -O2 -g -DNDEBUG -fPIC  -Wno-deprecated-objc-isa-usage -Wno-objc-root-class -fobjc-runtime=gnustep-2.0 -Xclang -x -Xclang assembler-with-cpp -o CMakeFiles/objc.dir/eh_trampoline.s.o -c eh_trampoline.s
```

triplef:
Ah I think the command for generating the assembly is missing a bunch of parameters such as `--target=armv7-none-linux-androideabi21`, so it’s generating assembly for a different ABI. I’ll try to figure out how to fix that...

davidchisnall:
Hmm.  That may have been broken by not passing `CMAKE_CXX_FLAGS` in that line, but we needed to make that change to avoid another problem.

triplef:
Yeah I was thinking that too (#178), but for me `CMAKE_CXX_FLAGS` doesn’t actually contain the --target, -gcc-toolchain, and --sysroot flags, which I think are what are needed. Where would these be defined? CXX_FLAGS does contain `-march=armv7-a`, but when I add that without the --target and --sysroot I get this error:
> error: unknown target CPU 'armv7-a'

Not sure if there’s a good way to get all the flags needed for cross compiling?

triplef:
After sifting through CMake variables I came up with something that seems to work ok. There are some additional flags in `CMAKE_CXX_FLAGS` such as `-march=armv7-a` and `-mthumb` which we would probably want in an ideal world, but it also seems to be ok to ignore these.

I also noticed that the build command for eh_trampoline.s was missing the `-Xclang -x -Xclang assembler-with-cpp` flags, which was caused by `set_source_files_properties()` being called before appending the file to `libobjc_ASM_SRCS`. While this didn’t cause any issues for me specifically I took the liberty of moving these calls to right before `add_library()` to ensure all files use the same flags.

davidchisnall:
> After sifting through CMake variables I came up with something that seems to work ok. There are some additional flags in `CMAKE_CXX_FLAGS` such as `-march=armv7-a` and `-mthumb` which we would probably want in an ideal world, but it also seems to be ok to ignore these.

These normally come from your toolchain file, but I don't know where it puts them.  Often these days, you'll provide a clang config file and then just set CC / CXX to a clang symlink with a prefix that matches your config file, but I don't know if Android does this.

> I also noticed that the build command for eh_trampoline.s was missing the `-Xclang -x -Xclang assembler-with-cpp` flags, which was caused by `set_source_files_properties()` being called before appending the file to `libobjc_ASM_SRCS`. While this didn’t cause any issues for me specifically I took the liberty of moving these calls to right before `add_library()` to ensure all files use the same flags.

That's only necessary on Windows, where we're not building these files, so it shouldn't make a difference (all of that can go away once we drop the requirement to build with clang-cl, but that needs a newer CMake).

triplef:
> > After sifting through CMake variables I came up with something that seems to work ok. There are some additional flags in `CMAKE_CXX_FLAGS` such as `-march=armv7-a` and `-mthumb` which we would probably want in an ideal world, but it also seems to be ok to ignore these.
> 
> These normally come from your toolchain file, but I don't know where it puts them. Often these days, you'll provide a clang config file and then just set CC / CXX to a clang symlink with a prefix that matches your config file, but I don't know if Android does this.

Thanks. I added these two using `CMAKE_SYSTEM_PROCESSOR` and `CMAKE_ANDROID_ARM_MODE`. There are still some others in CXX flags that affect ASM generation, but I think we can do without those as to not make this even less pretty:
`-g -fdata-sections -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes`

> > I also noticed that the build command for eh_trampoline.s was missing the `-Xclang -x -Xclang assembler-with-cpp` flags, which was caused by `set_source_files_properties()` being called before appending the file to `libobjc_ASM_SRCS`. While this didn’t cause any issues for me specifically I took the liberty of moving these calls to right before `add_library()` to ensure all files use the same flags.
> 
> That's only necessary on Windows, where we're not building these files, so it shouldn't make a difference (all of that can go away once we drop the requirement to build with clang-cl, but that needs a newer CMake).

Sounds good, I removed that change in order to keep this patch minimal.

triplef:
So unfortunately 7914c57 did not fix this for us, because we are still missing the `--target`, `--toolchain`, `--sysroot`, and `-march` parameters for cross compiling.

I’ve updated the patch based on the latest changes. It changes the eh_trampoline.s generation e.g. for Android x86 from this:

```
clang++ -g -DANDROID -fdata-sections -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -mstackrealign -D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fPIC -S eh_trampoline.cc -o - -fexceptions -fno-inline | sed s/__gxx_personality_v0/test_eh_personality/g > eh_trampoline.s
```

To this:
```
clang++ -g -DANDROID -fdata-sections -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -mstackrealign -D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security --target=i686-none-linux-android23 --gcc-toolchain=~/Library/Android/sdk/ndk/21.3.6528147/toolchains/llvm/prebuilt/darwin-x86_64 --sysroot=~/Library/Android/sdk/ndk/21.3.6528147/toolchains/llvm/prebuilt/darwin-x86_64/sysroot -march=i686 -fPIC -S eh_trampoline.cc -o - -fexceptions -fno-inline | sed s/__gxx_personality_v0/test_eh_personality/g > eh_trampoline.s
```

I also removed the `EH_PERSONALITY_M32` variable in the custom command, since it was seemingly never set.

davidchisnall:
Hmm, why are those parameters not in `CMAKE_CXX_FLAGS`?  How do they get into the compile commands for C++ files?  I'd expect a CMake cross-compile toolchain to add the -target, --sysroot and -B flags to `CMAKE_CXX_FLAGS`.

triplef:
I’m not so familiar with what CMake puts into `CMAKE_CXX_FLAGS`, but here’s the CMake invocation with the GNUstep Android toolchain (for x86):

```
cmake ..
    -DCMAKE_BUILD_TYPE=RelWithDebInfo
    -DCMAKE_TOOLCHAIN_FILE=~/Library/Android/sdk/ndk/21.3.6528147/build/cmake/android.toolchain.cmake
    -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=gold
    -DANDROID_ABI=x86
    -DANDROID_NDK=~/Library/Android/sdk/ndk/21.3.6528147
    -DANDROID_PLATFORM=android-23
    -DANDROID_STL=c++_shared
    -DCMAKE_INSTALL_PREFIX=~/Library/Android/GNUstep/x86
    -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=BOTH
    -DGNUSTEP_CONFIG=
    -DOLDABI_COMPAT=false
```

Maybe this is specific to how the [android.toolchain.cmake file](https://android.googlesource.com/platform/ndk/+/master/build/cmake/android.toolchain.cmake) from the NDK is set up?

triplef:
Looks like the toolchain file sets `CMAKE_SYSROOT`:
https://android.googlesource.com/platform/ndk/+/master/build/cmake/android.toolchain.cmake#408

davidchisnall:
I don't see where -target is passed, but it looks as if this toolchain file is setting most things in `CMAKE_CXX_FLAGS` (line 464).  If there are other flags that CMake passes into C++ compiler invocations, we should just unconditionally add them to the EH flags.

triplef:
Ok makes sense. I’ve updated the patch to always add these flags, let’s see what the CI says...

I’m not quite sure what to do about `-march`, because `CMAKE_SYSTEM_PROCESSOR` is probably always set. Should that one only be set if cross-compiling?

Btw. --target is set in the toolchain file like this:
`set(CMAKE_CXX_COMPILER_TARGET ${ANDROID_LLVM_TRIPLE})`

davidchisnall:
We shouldn't actually need `--sysroot` because this file doesn't include anything.  We shouldn't care about `-mthumb` (this file is sufficiently small and generates about two instructions, so even if they're non-Thumb AArch32 instructions we won't see a noticeable code size difference).  I don't think we care too much about `-march`, because we should get accurate enough information from the target.  `CMAKE_CXX_COMPILER_TARGET` seems to be the only thing we should actually need.

Given that this is specifically for the Android cross-compile environment, I'd like to minimise the maintenance burden in our CMake files.

triplef:
Nice, that seems to work just fine indeed and I think it’s as minimal as possible now. Thanks for working through this with me.

davidchisnall (approved):

davidchisnall:
Looks great, thanks!

PR #184, opened 2020-11-11
add 4 bytes of padding for stack alignment, as suggested by David.

rmottola:
Fixes start of most common applications.
Does not improve test cases on my system.


davidchisnall:
It looks as if we only run the 32-bit tests on Windows.  We should probably add something similar on Linux and FreeBSD.

rmottola:
I did run the test suite on FreeBSD x86-32 and all tests were passing before as well as after the patch. Running GUI apps was fine before and continues to be.

davidchisnall (approved):

PR #185, opened 2020-11-19
Increase testing matrix.

davidchisnall:
Add 32-bit x86-Linux builds.
Add building with the nightly LLVM releases.  There are some regressions
since LLVM 8 that would have been caught by this that we can hopefully
fix in time for the LLVM 12 release.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,38 +5,104 @@ jobs:
       vmImage: ubuntu-18.04
     strategy:
       matrix:
-        Debug:
+        Debug-64-llvm-8-libc++:
           BuildType: Debug
-          StdLib: libstdc++
-          CxxFlags: ""
-        Release:
+          Flags: -m64
+          StdLib: libc++
+          LLVM: -8
+        Release-64-llvm-8-libc++:
           BuildType: Release
-          StdLib: libstdc++
-          CxxFlags: ""
-        Debug-libc++:
+          Flags: -m64
+          StdLib: libc++
+          LLVM: -8
+        Debug-32-llvm-8-libc++:
           BuildType: Debug
+          Flags: -m32
           StdLib: libc++
-          CxxFlags: "-stdlib=libc++"
-        Release-libc++:
+          LLVM: -8
+        Release-32-llvm-8-libc++:
           BuildType: Release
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
Using different setups for the testing is in itself great. Fixing the issues that show up in this process is a different task.

triplef:
This is great. With the LLVM nightlies changing independently of this, would it make sense / be possible to set up an automated build to be run e.g. weekly to test the latest releases?

davidchisnall:
Yes, I'd like to make this do a weekly run.

davidchisnall:
A fix for the failing case is under review:

https://reviews.llvm.org/D91874

Once this is merged and the LLVM nightly build is updated, I'll merge this.

triplef:
I saw that Azure Pipelines recently added support for Linux ARM64 - would that make it easy to build and run the test for ARM 32 and 64 as well?

https://azure.microsoft.com/en-gb/updates/azure-devops-pipelines-introduces-support-for-linuxarm64/

davidchisnall:
Looks as if they've added Linux support for the pipelines agent (which basically means Arm Linux is now a supported platform for .NET Core), they haven't added hosted pipelines, so someone would need to stand up some Linux VMs to do that.  Using qemu user mode on x86 is probably easier.

davidchisnall:
New version is rebased on top of @ngrewe's changes, so tests every combination of:

- 32 / 64-bit
- Debug / Release
- libstdc++ / libc++ (this tests libsupc++ / lib++abi, the FreeBSD runners test the released and trunk versions of libcxxrt)
- Clang 8 / Clang nightly build

Each of these runs takes about 1.5 minutes.  We get 10 runners for free with Azure Pipelines and are building 20 configurations (16 Linux, 4 Windows).  The longest Windows runs take almost 6 minutes and so the 16 Linux builds on the 6 other runners don't take much longer.  It looks as if the Azure jobs still complete faster than the Cirrus CI ones, so this isn't adding anything to our total time.

I'd like to add Arm builds with the tests in qemu user mode at some point.  That will definitely slow things down, but given the commit rate in this repo even a 15-30 minute delay for the last tests in a PR is probably fine.

PR #186, opened 2020-11-28
Fix linker errors with GNUstep on Windows

triplef:
This fixes almost all of the [linker errors mentioned on the mailing list](https://mail.gnu.org/archive/html/gnustep-dev/2020-11/msg00018.html) when building GNUstep Base on Windows using Clang by:

- Exporting encoding functions.
- Fixing visibility of hooks by ensuring dllimport is used when building hooks.c.

The only remaining issue is the following, for which I have no idea where that symbol even comes from (as far as I understand this personality function should not be used on Windows):
```
lld-link: error: undefined symbol: __gnustep_objc_personality_v0
>>> referenced by obj/libgnustep-base.obj/GSICUString.m.o:(.xdata)
>>> referenced by obj/libgnustep-base.obj/NSRegularExpression.m.o:(.xdata)
>>> referenced by obj/libgnustep-base.obj/NSRegularExpression.m.o:(GCC_except_table11)
>>> referenced 1 more times
```

The only thing these files seem to have in common is that they both include `GSICUString.h` and some ICU headers, but when I build without ICU I get this:

```
lld-link: error: undefined symbol: objc_begin_catch
>>> referenced by obj/libgnustep-base.obj/GSHTTPAuthentication.m.o:(_c_GSHTTPAuthentication__authenticationWithCredential_inProtectionSpace_)
>>> referenced by obj/libgnustep-base.obj/GSHTTPAuthentication.m.o:(_c_GSHTTPAuthentication__protectionSpaceForURL_)
>>> referenced by obj/libgnustep-base.obj/GSHTTPAuthentication.m.o:(_c_GSHTTPAuthentication__setProtectionSpace_forDomains_baseURL_)
>>> referenced 134 more times

lld-link: error: undefined symbol: objc_end_catch
>>> referenced by obj/libgnustep-base.obj/GSHTTPAuthentication.m.o:(_c_GSHTTPAuthentication__authenticationWithCredential_inProtectionSpace_)
>>> referenced by obj/libgnustep-base.obj/GSHTTPAuthentication.m.o:(_c_GSHTTPAuthentication__authenticationWithCredential_inProtectionSpace_)
>>> referenced by obj/libgnustep-base.obj/GSHTTPAuthentication.m.o:(_c_GSHTTPAuthentication__protectionSpaceForURL_)
>>> referenced 251 more times

lld-link: error: undefined symbol: __gnustep_objc_personality_v0
>>> referenced by obj/libgnustep-base.obj/GSHTTPAuthentication.m.o:(.xdata)
>>> referenced by obj/libgnustep-base.obj/GSHTTPAuthentication.m.o:(GCC_except_table2)
>>> referenced by obj/libgnustep-base.obj/GSHTTPAuthentication.m.o:(GCC_except_table4)
>>> referenced 117 more times
```

Am I missing some flag to tell Clang to use MSVC EH? I also tried disabling native exceptions in Make with no change. (Sorry if this is all not strictly related to this PR.)

### DIFF
--- a/encoding2.c
+++ b/encoding2.c
@@ -19,6 +19,7 @@ size_t objc_alignof_type (const char *type);
 //
 #define alignof(type) __builtin_offsetof(struct { const char c; type member; }, member)
 
+OBJC_PUBLIC
 const char *objc_skip_type_qualifiers (const char *type)
 {
 	static const char *type_qualifiers = "rnNoORVA";
@@ -31,12 +32,14 @@ const char *objc_skip_type_qualifiers (const char *type)
 
 static const char *sizeof_type(const char *type, size_t *size);
 
+OBJC_PUBLIC
 const char *objc_skip_typespec(const char *type)
 {
 	size_t ignored = 0;
 	return sizeof_type(type, &ignored);
 }
 
+OBJC_PUBLIC
 const char *objc_skip_argspec(const char *type)
 {
 	type = objc_skip_typespec(type);
@@ -359,6 +362,7 @@ static const char *alignof_type(const char *type, size_t *align)
 	return NULL;
 }
 
+OBJC_PUBLIC
 size_t objc_sizeof_type(const char *type)
 {
 	size_t size = 0;
@@ -366,6 +370,7 @@ size_t objc_sizeof_type(const char *type)
 	return size / 8;
 }
 
+OBJC_PUBLIC
 size_t objc_alignof_type (const char *type)
 {
 	size_t align = 0;
@@ -373,6 +378,7 @@ size_t objc_alignof_type (const char *type)
 	return align / 8;
 }
 
+OBJC_PUBLIC
 size_t objc_aligned_size(const char *type)
 {
 	size_t size  = objc_sizeof_type(type);
@@ -380,12 +386,14 @@ size_t objc_aligned_size(const char *type)
 	return size + (size % align);
 }
 
+OBJC_PUBLIC
 size_t objc_promoted_size(const char *type)
 {
 	size_t size = objc_sizeof_type(type);
 	return size + (size % sizeof(void*));
 }
 
+OBJC_PUBLIC
 void method_getReturnType(Method method, char *dst, size_t dst_len)
 {
 	if (NULL == method) { return; }
@@ -403,13 +411,14 @@ void method_getReturnType(Method method, char *dst, size_t dst_len)
 	}
 }
 
+OBJC_PUBLIC
 const char *method_getTypeEncoding(Method method)
 {
 	if (NULL == method) { return NULL; }
 	return sel_getType_np(method->selector);
 }
 
-
+OBJC_PUBLIC
 void method_getArgumentType(Method method,
                             unsigned int index,
                             char *dst,
@@ -437,6 +446,7 @@ void method_getArgumentType(Method method,
 	}
 }
 
+OBJC_PUBLIC
 unsigned method_getNumberOfArguments(Method method)
 {
 	if (NULL == method) { return 0; }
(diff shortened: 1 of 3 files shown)


davidchisnall:
Are you using an msvc target triple for Windows?

davidchisnall (approved):

davidchisnall:
The Win32 failing tests is an intermittent failure that happens when the VM runs out of memory.

triplef:
> Are you using an msvc target triple for Windows?  

Do you mean for libobjc or GNUstep? I didn't manually specify a target triple for either, but I guess GNUstep makefiles might be choosing a wrong one.

I'll dig around for this. Can you tell me what we should be using?

davidchisnall:
I can't remember the exact one for Windows, but for the visual studio ABI, we need msvc in the triple.

triplef:
Hmm is it possible to use the MSVC ABI but still build using MinGW-w64 somehow, which I think is required for building GNUstep?

By default it seems to be using `x86_64-w64-mingw32`. I tried specifying some others like `x86_64-pc-windows-msvc` when building Make, but these are not recognized, which I’m guessing is expected behavior in a MinGW64 shell.

I might be missing something but right now I don’t see how GNUstep and libobjc2 fit together with these distinct requirements.

davidchisnall:
I think `x86_64-pc-windows-msvc` sounds right.  What error do you get?

triplef:
This is what I get for Make in a MinGW64 shell:

```
$ ./configure --prefix=/c/GNUstep/x64 --with-library-combo=ng-gnu-gnu --with-runtime-abi=gnustep-2.0 --target=x86_64-pc-windows-msvc LDFLAGS="-fuse-ld=lld"
configure: loading site script /mingw64/etc/config.site
checking build system type... x86_64-w64-mingw32
checking host system type... x86_64-w64-mingw32
checking target system type... Invalid configuration `x86_64-pc-windows-msvc': machine `x86_64-pc-windows' not recognized
configure: error: /bin/sh ./config.sub x86_64-pc-windows-msvc failed
```

I’m using the latest Clang from MSYS2 packages (`mingw-w64-x86_64-clang`):

```
$ clang -v
clang version 11.0.0 (https://github.com/msys2/MINGW-packages c0083b9edc111315c868f9c53dd60aaa78d1669b)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: C:\msys64\mingw64\bin
```

davidchisnall:
It looks as if you're passing the `--target` flag to `configure` and not to clang (via `CFLAGS` and friends)?

triplef:
Ah, yes that’s what I assumed was the thing to do. Passing --target via CFLAGS results the following:
```
$ ./configure --prefix=/c/GNUstep/x64 --with-library-combo=ng-gnu-gnu --with-runtime-abi=gnustep-2.0 CFLAGS=--target=x86_64-pc-windows-msvc LDFLAGS="-fuse-ld=lld"
configure: loading site script /mingw64/etc/config.site
checking build system type... x86_64-w64-mingw32
checking host system type... x86_64-w64-mingw32
checking target system type... x86_64-w64-mingw32
checking for pkgconfig... yes
checking for gnustep-config... no
checking for gawk... gawk
checking for library combo... ng-gnu-gnu
checking for gcc... clang
checking whether the C compiler works... no
configure: error: in `/c/Dev/tools-android/src/gnustep-make':
configure: error: C compiler cannot create executables
See `config.log' for more details
```
config.log reveals that the MinGW clang is missing the VC libs:
```
configure:3163: clang --target=x86_64-pc-windows-msvc  -fuse-ld=lld conftest.c  >&5
lld-link: error: could not open 'libcmt.lib': No such file or directory
```

So I tried passing the VC libraries directory (`C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.28.29333\lib\x64`, using a symlink to get rid of the spaces) via LDFLAGS, but then it fails a bit later trying to find stdio.h.

I also tried using the pre-built Clang release from the LLVM website, which defaults to the `x86_64-pc-windows-msvc` target. This works fine for building Make, but then building Base fails with:
`configure: error: Unable to find pthread.h (needed for thread support).`

I’m guessing that’s because this is not providing the MinGW environment needed for Base?

PR #188, opened 2020-12-15
tests: Test that we can throw and catch the same exception multiple times sequentially.

triplef:
This originally came up as an issue with libc++abi support (#152), but is not specific to that ABI.

It was extracted from the `libc++` branch as to unblock merging #152.

The test currently fails, but the solution was discussed in https://github.com/gnustep/libobjc2/pull/152#issuecomment-694906062.


davidchisnall:
Rebased on master and with the fix from #152 added.

triplef:
Nice, thank you!

ngrewe (approved):

triplef:
I rebased to fix a small merge conflict. I think we can merge this @davidchisnall?

PR #194, opened 2021-02-18
Remove redundant declaration of __cxa_allocate_exception

### LINKED ISSUE #193: Compile error with clang on armbian aarch64
## Environment

clang 3.8.1, cmake 3.15.4, on Armbian (Debian 9.13)

## Steps to reproduce

```
export CC=clang
export CXX=clang++
cmake ..
make -j4
```

## Error message

```
[  7%] Building CXX object CMakeFiles/objc.dir/objcxx_eh.cc.o
/home/leeg/Projects/libobjc2/objcxx_eh.cc:19:18: error: exception specification
      in declaration does not match previous declaration
extern "C" void *__cxa_allocate_exception(size_t) noexcept;
                 ^
/home/leeg/Projects/libobjc2/objcxx_eh.h:18:7: note: previous declaration is
      here
void *__cxa_allocate_exception(size_t thrown_size) CXA_ALLOCATE_EXCEPTIO...
      ^
1 error generated.
```

## What I tried

Installing llvm libc++, and using ccmake to `-DCXX_RUNTIME_LIB=/usr/lib/aarch64-linux-gnu/libc++.so` yielded this error:

```
 CMake Error at CMakeLists.txt:321 (separate_arguments):
   separate_arguments given unknown argument NATIVE_COMMAND
```

So I then just tried passing that constructed argument to cmake. That worked, but I get the same error.

## Notes

In fairness the comment right above this line says that clang would throw a type mismatch error, I would switch to gcc except the cmake configure phase strongly advises against that:

> WARNING: It is strongly recommended that you compile with clang

iamleeg:
Fixes #193

### DIFF
--- a/objcxx_eh.cc
+++ b/objcxx_eh.cc
@@ -16,8 +16,6 @@ int eh_trampoline();
 
 uint64_t cxx_exception_class;
 
-extern "C" void *__cxa_allocate_exception(size_t) noexcept;
-
 /**
  * Our own definitions of C++ ABI functions and types.  These are provided
  * because this file must not include cxxabi.h.  We need to handle subtly


davidchisnall:
Thanks!

davidchisnall (approved):

PR #200, opened 2021-03-14
Avoid dangling weak references to deallocating objects.

### LINKED ISSUE #198: Possible dangling pointer when loading a weak reference about to deallocate?
I'm having a some trouble comprehending the reference counting code in `arc.mm`.

As I understand it:
- The most significant bit/sign bit is used to indicate that a weak reference has been taken.
- The other bits are for the reference count.
- Instances start with a reference count of 0 (which indicates the instance has only one strong reference).
- When all the reference count bits are set, this means that the reference count has saturated...
- ...or when releasing from a reference count of 0, the underflow indicates the instance should be destroyed.
- When the reference count has saturated, further retains and releases are ignored and the instance has a permanent lifetime.

My main concern is in `objc_retain_fast_np()` where [here](https://github.com/gnustep/libobjc2/blob/c399119694ae91b02286ecacfed1f4826bbc99e5/arc.mm#L271), it tries to handle the case where one thread has done the final release of a strong reference but before it can grab the weak reference lock to remove the `WeakRef` from the map and set the pointer to `nil`, another thread attempts to load and retain the instance through the weak reference.

I can't see how `realCount < 0` can ever happen because the sign bit is masked off. When I disassemble the function in a release build, that branch has been removed by the optimisation:

```
(gdb) disassemble objc_retain_fast_np
Dump of assembler code for function objc_retain_fast_np:
   0x00007ffff7fb5ad0 <+0>:     movabs r8,0x7fffffffffffffff
   0x00007ffff7fb5ada <+10>:    mfence 
   0x00007ffff7fb5add <+13>:    mov    rax,QWORD PTR [rdi-0x8]
   0x00007ffff7fb5ae1 <+17>:    lea    rdx,[r8+0x1]
   0x00007ffff7fb5ae5 <+21>:    nop    WORD PTR cs:[rax+rax*1+0x0]
   0x00007ffff7fb5aef <+31>:    nop
   0x00007ffff7fb5af0 <+32>:    mov    rsi,rax
   0x00007ffff7fb5af3 <+35>:    and    rsi,r8
   0x00007ffff7fb5af6 <+38>:    cmp    rsi,r8
   0x00007ffff7fb5af9 <+41>:    je     0x7ffff7fb5b10 <objc_retain_fast_np+64>
   0x00007ffff7fb5afb <+43>:    add    rsi,0x1
   0x00007ffff7fb5aff <+47>:    mov    rcx,rax
   0x00007ffff7fb5b02 <+50>:    and    rcx,rdx
   0x00007ffff7fb5b05 <+53>:    or     rcx,rsi
   0x00007ffff7fb5b08 <+56>:    lock cmpxchg QWORD PTR [rdi-0x8],rcx
   0x00007ffff7fb5b0e <+62>:    jne    0x7ffff7fb5af0 <objc_retain_fast_np+32>
   0x00007ffff7fb5b10 <+64>:    mov    rax,rdi
   0x00007ffff7fb5b13 <+67>:    ret    
End of assembler dump.
```
So it looks like the retain always returns a pointer to the instance. `setObjectHasWeakRefs()` has a similar logic.

I'm also wondering how the difference between a saturated count and an instance deallocating is handled. If you handle the case where the instance is deallocating, wouldn't it mean that you can never load and retain a weak pointer to an instance that has a saturated retain count?

Graham--M:
Fixes #198

The previous checks for a deallocating object with negative reference count did not work because:
    1. the sign bit has to be recreated as was happening in objc_delete_weak_refs()
    2. there was no distinction between a saturated count and a negative reference count

refcount_max now indicates when the refcount has saturated and should no longer be mutated.
An underflow to -1 still maps to refcount_mask, allowing us to detect when an object is supposed to be deallocated.
Neither objc_release_fast_no_destroy_np() nor objc_retain_fast_np() mutate the refcount when it is one of those values, so the comment in objc_delete_weak_refs() was adjusted.

### DIFF
--- a/Test/CMakeLists.txt
+++ b/Test/CMakeLists.txt
@@ -27,6 +27,7 @@ set(TESTS
 	Category.m
 	ExceptionTest.m
 	FastARC.m
+	FastRefCount.m
 	Forward.m
 	ManyManySelectors.m
 	NestedExceptions.m
--- a/Test/FastRefCount.m
+++ b/Test/FastRefCount.m
@@ -1,0 +1,149 @@
+#include "Test.h"
+
+void direct_saturation_test();
+
+int main()
+{
+	id obj = [Test new];
+	assert(object_getRetainCount_np(obj) == 1);
+	
+	for (int i = 0; i < 2; i++)
+	{
+		size_t count = object_getRetainCount_np(obj);
+		id ret = objc_retain_fast_np(obj);
+		assert(ret == obj);
+		assert(object_getRetainCount_np(obj) == ++count);
+	}
+	
+	for (int i = 0; i < 2; i++)
+	{
+		size_t count = object_getRetainCount_np(obj);
+		BOOL destroy = objc_release_fast_no_destroy_np(obj);
+		assert(destroy == NO);
+		assert(object_getRetainCount_np(obj) == --count);
+	}
+	
+	{
+		// Final release should prevent further retains and releases.
+		assert(objc_release_fast_no_destroy_np(obj) == YES);
+		assert(object_getRetainCount_np(obj) == 0);
+		assert(objc_retain_fast_np(obj) == nil);
+		assert(object_getRetainCount_np(obj) == 0);
+		assert(objc_release_fast_no_destroy_np(obj) == NO);
+		assert(object_getRetainCount_np(obj) == 0);
+	}
+	
+	object_dispose(obj);
+	obj = [Test new];
+	
+	{
+		// Should not be able to delete weak refs until final release.
+		id weak;
+		assert(objc_initWeak(&weak, obj) == obj);
+		assert(weak != nil);
+		assert(objc_loadWeakRetained(&weak) == obj);
+		assert(objc_release_fast_no_destroy_np(obj) == NO);
+		// Assumes a return of NO means no effect on obj at all.
+		assert(objc_delete_weak_refs(obj) == NO);
+		assert(objc_loadWeakRetained(&weak) == obj);
+		assert(objc_release_fast_no_destroy_np(obj) == NO);
+		// This will also call objc_delete_weak_refs() and succeed.
+		assert(objc_release_fast_no_destroy_np(obj) == YES);
+		objc_destroyWeak(&weak);
+		// Check what happens when the weak refs were already deleted.
+		assert(objc_delete_weak_refs(obj) == YES);
+	}
+	
+	obj = object_dispose(obj);
+	
+	direct_saturation_test();
+	return 0;
+}
+
+
+// ----------------
+// This test has knowledge of the implementation details of the ARC
+// reference counting and may need modification if the details change.
+
+const long refcount_shift = 1;
+const size_t weak_mask = ((size_t)1)<<((sizeof(size_t)*8)-refcount_shift);
+const size_t refcount_mask = ~weak_mask;
+const size_t refcount_max = refcount_mask - 1;
+
+size_t get_refcount(id obj)
+{
+	size_t *refCount = ((size_t*)obj) - 1;
+	return *refCount & refcount_mask;
+}
+
+void set_refcount(id obj, size_t count)
+{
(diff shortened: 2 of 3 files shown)


Graham--M:
Hmm. The `Windows-2016 Debug-32` job seems to be stuck on the `ManyManySelectors` test.
Every other job completed under 6 minutes.

davidchisnall:
The Win32 builds sometimes run out of memory on those tests, I should probably disable them there.  

Would you mind adding a test so that we catch regressions here?

Graham--M:
Sure, I'll give it a shot. Presumably it's acceptable to directly set the reference count in order to check the behaviour?

Graham--M:
I've added a test for the expected behaviours. There is a little overlap with `WeakRefLoad.m` but I think it's better to keep the simple checks in that test.

The pre-emption case can't be directly tested for in a reasonable manner but it's less likely to happen if the functions involved pass the testing. The behaviour around the count saturating is tested by directly setting the refcount to avoid doing a ridiculous number of retains.

I went to add a check that you can't init a weak ref to a deallocating object then load it but then I noticed that `objc_initWeak()` and `objc_storeWeak()` are supposed to set the pointer to `nil` if the object is deallocating, but it's currently possible to init a weak ref when this is the case.

Should I try handle this? It would mean either moving the persistent object check out of `setObjectHasWeakRefs()` or returning the deallocation status from it by reference.

davidchisnall (approved):
Great, thanks!

triplef:
I think this change may have caused the following failures in GNUstep Base tests:
```
NSFileHandle/singleton:
Failed test:     (2021-03-17 16:59:18.408 +0000) singleton.m:24 ... stdin persists
Failed test:     (2021-03-17 16:59:18.408 +0000) singleton.m:25 ... stdout persists
Failed test:     (2021-03-17 16:59:18.408 +0000) singleton.m:26 ... strerr persists
```

`GSFileHandle` will resurrect objects by calling -retain in -dealloc:
https://github.com/gnustep/libs-base/blob/753c907938c2a8c4d00cf0fbe01b7e0d020f0064/Source/GSFileHandle.m#L354

Could this have been affected by this change?

davidchisnall:
@triplef, would you be able to prepare a reduced test case for this?

triplef:
I tried, but unfortunately wasn’t able to repro this in a straightforward way in libobjc2 tests. I’m guessing that’s because Base overriding retain/release/retainCount in NSObject is affecting this, but I haven’t figured out how exactly.

davidchisnall:
Looking at the GNUstep code, it appears that it's relying on undefined behaviour: Resurrection is not permitted in Objective-C.  I don't know what macOS does if you try to deallocate a singleton, I'd imagine that it aborts the program.  With ARC and weak references, the weak references should have all gone away by the time that -dealloc is called and so if you'll end up zeroing the weak references but not destroying the object, which is a very odd behaviour.

triplef:
That makes sense. I already [changed](https://github.com/gnustep/libs-base/commit/b05481a8c89c80d28b6c15907db52c9cd0667379) the GSFileHandle implementation to no longer resurrect in dealloc.

However, I think I found another issue with this change: `self` no longer works inside a block that is run from within `dealloc` when building with ARC.

```objc
- (void)ping
{
  NSLog(@"%@ ping", self);
}

- (void)dealloc
{
  [self ping];

  void (^block)() = ^{
    NSLog(@"%@ block", self);
    [self ping];
  };
  block();
}
```

The above will print the following (note that there should be a second "ping"):
```
<Test: 0xb855aef4> ping
(null) block
```

We found this because YapDatabase is doing [things](https://github.com/yapstudios/YapDatabase/blob/master/YapDatabase/YapDatabaseConnection.m#L407-L427) along the lines of the above.

Graham--M:
>However, I think I found another issue with this change: self no longer works inside a block that is run from within dealloc when building with ARC.

I'll try make a PR that always returns the instance pointer when it's not a retain for a weak reference.

PR #202, opened 2021-03-26
Export objc_enumerationMutation() function

triplef:
Fixes linker errors on Windows when using for-in loops.

### DIFF
--- a/mutation.m
+++ b/mutation.m
@@ -4,6 +4,7 @@
 
 // This function is exported as a weak symbol to enable GNUstep or some other
 // framework to replace it trivially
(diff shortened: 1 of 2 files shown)


davidchisnall (approved):

davidchisnall:
I don't believe that weak symbols are not preemptable (or, in fact, supported) on Windows, so for Windows we probably need to:

 - Make this invoke a callback.
 - Add a function to register the callback.

triplef:
Good to know. This is in fact how it’s done on Apple platforms via `objc_setEnumerationMutationHandler()`.

PR #204, opened 2021-04-08
Fix retain calls in dealloc methods to return the instance pointer as expected

Graham--M:
This addresses a regression in 73132a6 (#200) where retains of an object inside its `dealloc` method no longer return a pointer to the instance and this behaviour prevents blocks inside a `dealloc` from capturing the `self` pointer via `objc_storeStrong()`.

The `objc_retain(id value)` method is defined in the Clang ARC ABI supplement to always return the value passed into it.

### DIFF
--- a/Test/FastRefCount.m
+++ b/Test/FastRefCount.m
@@ -2,6 +2,20 @@
 
 void direct_saturation_test();
 
+@interface TestWithDelloc : Test
+@end
+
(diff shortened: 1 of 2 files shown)


davidchisnall (approved):

triplef:
Thanks for fixing this @Graham--M!

PR #206, opened 2021-04-24
Updated the class_copyMethodList

mischievous:
Updated the class_copyMethodList to clear the stack garbage for outCount if preset before checking the class is null.  This allows the pyobjc to mostly work.

### DIFF
--- a/runtime.c
+++ b/runtime.c
@@ -208,6 +208,11 @@ BOOL class_addProtocol(Class cls, Protocol *protocol)
 
 Ivar * class_copyIvarList(Class cls, unsigned int *outCount)
 {
+    if (outCount != NULL)
+    {
+        *outCount = 0x0;
+    }
+
 	CHECK_ARG(cls);
 	struct objc_ivar_list *ivarlist = NULL;
 	unsigned int count = 0;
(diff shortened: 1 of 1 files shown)


davidchisnall:
Thanks.  This looks functionally correct, but please add the missing braces around the body of the if statement.

mischievous:
So the automatic testing for gnustep.libobjc2  claimed it took longer than 60 minutes and was canceled...   How do I resolve this issue?

davidchisnall:
Thanks.  There's an intermittent failure in one of the tests on 32-bit Windows.

mischievous:
Tell me where to look to fix the test, and/or how to test it again and I will get it fixed.

davidchisnall:
Sorry, to be clear: It's not your fault.  The ManyManySelectors test allocates a lot of memory and, for some reason that I haven't been able to determine, sometimes on Win32 it times out rather than failing.  I haven't been able to reproduce this, it seems to be something specific to how the Windows CI VMs work.  It's intermittent and just rerunning the timed-out job typically fixes it.  The timeout behaviour doesn't seem to happen anywhere else (this test can run out of memory on other systems, but then it fails).

PR #211, opened 2021-10-07
Set that same options on the `objc-static` target as the `objc` target

madsmtm:
I still can't get static linking to work, but this makes it slightly better


davidchisnall:
From the error, it look as if `libunwind` is being added on static builds now, but that functionality should come from something with a different name on static builds?  Or maybe Ubuntu 18.04's clang just doesn't ship a static libunwind?

davidchisnall:
Was this superseded by #380?

madsmtm:
I think so, yeah!

PR #220, opened 2021-12-22
Support _objc_unexpected_exception hook on Win32

### LINKED ISSUE #219: _objc_unexpected_exception hook not implemented on Windows
We’d like to use the `_objc_unexpected_exception` hook in order to send unhandled exceptions to our crash reporting including the exception reason for context. While this hook is handled for other platforms in `eh_personality.c`, the Windows implementation in `eh_win32_msvc.cc` seems to be lacking support for this.

Would there be a way to implement this in the runtime, or do we need to do what WinObjC is doing and use `SetUnhandledExceptionFilter()` in our code to install a [handler like this](https://github.com/microsoft/WinObjC/blob/develop/Frameworks/Foundation/NSException.cl.cpp)?

triplef:
This adds support for the `_objc_unexpected_exception` hook on Windows as discussed in #219.

GNUstep Base setting `_objc_unexpected_exception` worked fine in my testing and `callUncaughtHandler()` is called with these changes, contrary to what @davidchisnall you said about exposing global variables from a DLL. Not sure if I’m missing something there.

The only issue I ran into was that setting the exception filter via `SetUnhandledExceptionFilter()` from a constructor block didn’t work for me, I think because the VC runtime [sets a handler itself](https://github.com/ojdkbuild/tools_toolchain_vs2019bt_16113/blob/8739e474753c93215ecade7c36bb0f96b7f3903b/VC/Tools/MSVC/14.29.30133/crt/src/vcruntime/utility_desktop.cpp#L101) at startup, which at least for me was done after the constructor was called. I ended up setting it at the top of `objc_exception_throw()` instead, which I’m not really happy with unless we could restore the original handler afterwards, but I’m not sure if we get any callback if the exception is actually handled. I’d appreciate any thoughts here to improve this.

Fixes #219.


triplef:
I added a test for the `_objc_unexpected_exception` hook (passed on both 32- and 64-bit Windows).

eh_win32_msvc.cc
```diff
@@ -51,6 +56,8 @@ struct _MSVC_ThrowInfo
 	unsigned long pCatchableTypeArray;
 };
 
+extern "C" LONG WINAPI _objc_unhandled_exception_filter(struct _EXCEPTION_POINTERS* exceptionInfo);
```
davidchisnall, line 59: Does this actually need to be `extern`?  It shouldn't be referenced anywhere other than this compilation unit and then passed by a function pointer, so `WINAPI` ought to be sufficient.

eh_win32_msvc.cc
```diff
@@ -105,6 +112,9 @@ OBJC_PUBLIC extern "C" void objc_exception_rethrow(void* exc);
 
 OBJC_PUBLIC extern "C" void objc_exception_throw(id object)
 {
+	// Set unhandled exception filter to support _objc_unexpected_exception hook
+	SetUnhandledExceptionFilter(&_objc_unhandled_exception_filter);
```
davidchisnall, line 116: Do we need to do this every time an exception is thrown, or can we do it once on init time?  If we do need to do it here, can you add a comment explaining why please?

eh_win32_msvc.cc
```diff
@@ -203,10 +210,52 @@ OBJC_PUBLIC extern "C" void objc_exception_throw(id object)
...
-
 OBJC_PUBLIC extern "C" void objc_exception_rethrow(void* exc)
 {
 	_CxxThrowException(nullptr, nullptr);
 	__builtin_unreachable();
 }
 
+// rebase_and_cast adds a constant offset to a U value, converting it into a T
+template <typename T, typename U>
+static std::add_const_t<std::decay_t<T>> rebase_and_cast(size_t base, U value) {
```
davidchisnall, line 221: Please don't use `size_t` for values that carry pointer provenance.  `intptr_t` exists for that purpose.

eh_win32_msvc.cc
```diff
@@ -203,10 +210,52 @@ OBJC_PUBLIC extern "C" void objc_exception_throw(id object)
...
+	// U value -> const T* (base+value)
+	return reinterpret_cast<std::add_const_t<std::decay_t<T>>>(base + (long)(value));
+}
+
+extern "C" LONG WINAPI _objc_unhandled_exception_filter(struct _EXCEPTION_POINTERS* exceptionInfo)
```
davidchisnall, line 226: Please can you add a doc comment for this function, including a reference to the docs for people who want to understand how this interacts with the Win32 APIs?

davidchisnall (commented):
Thanks for doing this.  A few minor comments, but otherwise it looks good.

eh_win32_msvc.cc
triplef: Good point.

eh_win32_msvc.cc
triplef: As I mentioned in the issue description, I tried installing the handler in a constructor block (using `__attribute__((constructor))`), but this seems to get called before the VC runtime gets initialized, which installs its own hook and thus ours gets overwritten.

I’m not very happy calling this on every exception, as it means that throwing an exception has the side effect of overwriting any existing exception handler, and as far as I can see there’s no way for us to restore it as we can’t execute code after an exception is possibly handled. However I couldn’t come up with any other solution so this is what I went with.

For now I added a comment to explain the above, but if you can think of any other way to do this please let me know.

triplef: Since `SetUnhandledExceptionFilter()` returns the previous handler (if any), I now save that and call it as part of our handler. Does that sound ok to you?

The only issue I see with this is that if another piece of code would later try to unset the handler it had previously set (and which we stored) that would no longer work, but since handlers can get overwritten it seems like it’s something that any code using these handler should have to account for anyway.

triplef:
I found one strange issue while working on this further: when running any process throwing an _unhandled_ ObjC exception I always get an access violation like this (with varying addresses) after `RtlRaiseException()` is called. However when I run without a debugger this doesn’t seem to happen. It’s easily reproducible using the UnexpectedException test. Any idea what could be going on here? The issue seems to come from a std::string getting released (see second screenshot), although it seems to me like this one should be released before `RtlRaiseException()`?

image
image

eh_win32_msvc.cc
davidchisnall: That sounds sensible to me.

davidchisnall:
Not sure what's going on there.  I found trying to debug this code in Visual Studio often didn't work because it installed hooks into a bunch of places in the unwind machinery that caused it to not work properly.

triplef:
I rebased this to get the latest CI builds and the cross builds are all failing the new UnexpectedException test. Is `_objc_unexpected_exception` not supported on ARM? I thought it was.

triplef:
I was looking at this again to try to find a place where we can install our exception handler in a way so it doesn’t get overwritten by the CRT.

`SetUnhandledExceptionFilter()` is called as part of the runtime’s [`pre_cpp_initialization()`](https://github.com/ojdkbuild/tools_toolchain_vs2019bt_16113/blob/8739e474753c93215ecade7c36bb0f96b7f3903b/VC/Tools/MSVC/14.29.30133/crt/src/vcruntime/exe_common.inl#L215-L223) (installed via `.CRT$XCAA`), which unfortunately gets called after `__attribute__((constructor))` blocks and static struct constructors.

@davidchisnall can you think of another way we can ensure the CRT gets initialized before we install our handler? Here’s the stack trace leading up to setting the exception handler in the runtime:

```
	KernelBase.dll!SetUnhandledExceptionFilter()	Unknown
 	app.exe!__scrt_set_unhandled_exception_filter() Line 102	C++
 	app.exe!pre_cpp_initialization() Line 222	C++
 	ucrtbased.dll!_initterm(void(*)() * first, void(*)() * last) Line 22	C++
 	app.exe!__scrt_common_main_seh() Line 258	C++
 	app.exe!__scrt_common_main() Line 331	C++
 	app.exe!WinMainCRTStartup(void * __formal) Line 17	C++
 	kernel32.dll!BaseThreadInitThunk()	Unknown
 	ntdll.dll!RtlUserThreadStart()   Unknown
```

davidchisnall:
Have you looked at the way we register the calls to the entry points?  We put our init code in the `.CRT$XCLz` section.  The `.CRT` section contains function pointers to init functions and relies on the fact that subsections are sorted.  The XCL prefix is for libraries, which run before user code.  The z suffix means 'after all the others please'.  You can try sticking a global variable that points to a function that registers the hook in a global in a section with a name like this (use the section attribute) and see where it can live to not get clobbered?

triplef:
You mean like this?

```c
void _objc_install_unhandled_exception_filter(void)
{
	LPTOP_LEVEL_EXCEPTION_FILTER previousExceptionFilter = SetUnhandledExceptionFilter(&_objc_unhandled_exception_filter);
	if (previousExceptionFilter != &_objc_unhandled_exception_filter) {
		originalUnhandledExceptionFilter = previousExceptionFilter;
	}
}

__attribute__((section(".CRT$XCLz"), used)) static void (*__objc_init)(void) = _objc_install_unhandled_exception_filter;
```

Unfortunately that also gets called before pre_cpp_initialization().

davidchisnall:
I believe XCU is used for user code (not library) initialisation, so XCUa might work?

triplef:
That still gets called before `pre_cpp_initialization()` unfortunately. I also tried XCUz, and verified we don’t get called at all when I comment out the variable – just to make sure I’m not doing something silly.

davidchisnall:
What about ZZZz?

triplef:
ZZZz doesn’t seem to get called at all.

I’m not sure if this approach will work at all, since if I understand this correctly all of these pointers in the .CRT sections get called on DLL load, and for me ucrtbased.dll executing the code shown above gets loaded after objc.dll. So we would have to somehow delay setting the exception hander until after, not sure if that’s even possible?

That’s why I was wondering if we should expose a function for the user to call to register the handler. In fact I just realized that the Apple runtime has `objc_setUncaughtExceptionHandler()` instead of the `_objc_unexpected_exception` function pointer. What do you think about switching to that? GNUstep Base [supports both](https://github.com/gnustep/libs-base/blob/7fe39a23e9d9b3d09954020f694055b295c08cf4/Source/NSException.m#L1448) and will detect the availability during configure.

triplef:
I’ve implemented the `objc_setUncaughtExceptionHandler()` API, let me know what you think.

I left the hook defined for non-Windows platforms and just added a not about deprecation. Happy to remove it completely as well.

triplef:
I just realized that GNUstep Base [expects](https://github.com/gnustep/libs-base/blob/7fe39a23e9d9b3d09954020f694055b295c08cf4/config/config.setUncaughtExceptionHandler.m#L3) this new function to be in `objc/objc-exception.h` like in the Apple runtime. Would it make sense to add that header, or should I patch that config check instead (e.g. using __has_include)?

davidchisnall:
It’s a good idea to put it in the header.  Happy to remove the global unconditionally on Windows and deprecate it elsewhere.

triplef:
Sounds good, I’ll make that change.

What should we do about the crashes on the cross-build ARM CI targets – any idea what could be going on there? If it’s ok I could just disable the new test on ARM for now so we can merge this, and open a ticket about it. I’m curious about it because we do use the exception hook on Android ARM and it’s working fine for us there.

eh_personality.c
```diff
@@ -760,3 +761,10 @@ void objc_exception_rethrow(struct _Unwind_Exception *e)
 	_Unwind_Resume_or_Rethrow(e);
 	abort();
 }
+
+objc_uncaught_exception_handler objc_setUncaughtExceptionHandler(objc_uncaught_exception_handler handler)
+{
+	objc_uncaught_exception_handler previousHandler = _objc_unexpected_exception;
```
davidchisnall, line 767: This should be an atomic exchange.

davidchisnall:
> What should we do about the crashes on the cross-build ARM CI targets – any idea what could be going on there? If it’s ok I could just disable the new test on ARM for now so we can merge this, and open a ticket about it. I’m curious about it because we do use the exception hook on Android ARM and it’s working fine for us there

No, that's a bit worrying, since the test that's failing is for the functionality that you've modified.  Can you try debugging it locally?  That test was passing last time it ran on the main branch, so it looks like a regression.

triplef:
The failing "UnexpectedException" was added with this PR, I believe the tests just didn’t cover the exception hook before. Since I only touched the Windows implementation I don’t think it’s a new issue but I’ll try to figure out how to debug it locally.

eh_personality.c
triplef: Sure, would you use `atomic_exchange()` and declare `_objc_unexpected_exception` as volatile, or is there another way without needing volatile?

Btw. the Apple runtime doesn’t do an atomic exchange:
https://github.com/opensource-apple/objc4/blob/cd5e62a5597ea7a31dccef089317abb3a661c154/runtime/objc-exception.mm#L455

triplef: Sorry which macros exactly? Do you mean `objc_atomicCompareAndSwapPtr()`?

davidchisnall:
Huh, I thought we had an atomic.h that wrapped the builtins.  Maybe I just used the compiler builtins directly elsewhere since we can rely on being built with a clang-compatible compiler.

eh_personality.c
triplef: Sorry I’m not so familiar with builtins. Would the following make sense at all?
```
return __atomic_exchange(&_objc_unexpected_exception, handler, __ATOMIC_RELAXED);
```

triplef: I ended up using the following, I hope that makes sense?
```c
__atomic_exchange_n(&_objc_unexpected_exception, handler, __ATOMIC_SEQ_CST);
```

davidchisnall: Looks good.  We probably can get away with a slightly weaker memory order, but if anyone calls this enough times for it to matter for performance then they deserve for it to be slow.

triplef: Heh sounds good. I just copied this from somewhere... 🙈  Just let me know if I should change it in any way.

The test is also excluded on ARM now so CI is green.

davidchisnall (approved):

triplef:
Thanks for your help @davidchisnall!

davidchisnall:
Thanks for actually doing all of the work!  I just heckle...

PR #221, opened 2021-12-29
Various CMake cleanups and modernisations.

davidchisnall:


triplef:
Thanks! I’ll give this a try with our Android and Windows toolchains (probably the second week of January after the holidays).

Looks like the Windows builds are still failing though (and we might also want to upgrade the CI to VS 2019 or 2022 according to the warning).

CMakeLists.txt
```diff
@@ -193,37 +148,14 @@ set(INSTALL_TARGETS objc)
...
-
 #
 # C++ Runtime interaction
 #
 
 
 function(test_cxx CXX_RUNTIME_NAME IS_STDLIB)
 	set(CXX_RUNTIME_NAME_FULL "${CMAKE_SHARED_LIBRARY_PREFIX}${CXX_RUNTIME_NAME}${CMAKE_SHARED_LIBRARY_SUFFIX}")
-	find_library(CXX_RUNTIME_LIB NAMES ${CXX_RUNTIME_NAME_FULL} ${CXX_RUNTIME_NAME})
+	find_library(CXX_RUNTIME_LIB NAMES ${CXX_RUNTIME_NAME_FULL})
```
triplef, line 158: Why was this parameter removed? When building for Android, I previously got the following log output:
```
-- Testing C++ interop
-- Testing <ndk-path>/toolchains/llvm/prebuilt/darwin-x86_64/sysroot/usr/lib/arm-linux-androideabi/libc++abi.a as the C++ runtime library
-- Using <ndk-path>/toolchains/llvm/prebuilt/darwin-x86_64/sysroot/usr/lib/arm-linux-androideabi/libc++abi.a as the C++ runtime library
```

Building the branch will not find libc++abi.a, and instead go the "Testing C++ standard library" route.

triplef (changes requested):
I tried this building the project on macOS for Android and on Windows.

For Android, libc++abi is not being found because the marked change when calling `find_library()`.

For Windows I’m getting the following error (same as on CI I think):
```
CMake Error at C:/Program Files/Microsoft Visual Studio/2022/Professional/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.21/Modules/Compiler/CMakeCommonCompilerMacros.cmake:42 (message):
  CMAKE_OBJC_STANDARD_COMPUTED_DEFAULT should be set for Clang (C:/Program
  Files/Microsoft Visual
  Studio/2022/Professional/VC/Tools/Llvm/x64/bin/clang-cl.exe) version 12.0.0
Call Stack (most recent call first):
  C:/Program Files/Microsoft Visual Studio/2022/Professional/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.21/Modules/Compiler/Clang-OBJC.cmake:39 (__compiler_check_default_language_standard)
  C:/Program Files/Microsoft Visual Studio/2022/Professional/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.21/Modules/CMakeOBJCInformation.cmake:25 (include)
  CMakeLists.txt:3 (project)
```

Let me know if there’s anything else I can try.

triplef:
Talking to @weliveindetail today about extending libobjc2 tests (e.g. to cover additional EH edge cases that still have issues on Windows), we thought it would be helpful if the test suite also ran on Mac (excluding tests that cover libobjc2-specific behavior), in order to be able to compare and match behavior between the runtimes as much as possible. Do you think this would make sense @davidchisnall?

I’m adding this here, since this patch using native CMake ObjC support would probably be somewhat of a prerequisite for running the tests on Mac.

davidchisnall:
That was always one of my goals.  I used to run the tests manually on macOS.

triplef:
Can you share how you ran them manually on macOS?

triplef:
We’ve been looking into setting up CMake for our own projects and on Windows are running into the same CMake error mentioned above using CMake 3.24.2:
```
`CMAKE_OBJC_STANDARD_COMPUTED_DEFAULT` should be set for Clang
```

After looking at the [source of the error message in CMake](https://github.com/Kitware/CMake/blob/ff875ed8591afdb87055b8e43b1fcd04438a4d1e/Modules/Compiler/CMakeCommonCompilerMacros.cmake#L42), I was able to work around this as follows:

```cmake
foreach(lang IN ITEMS OBJC OBJCXX)
    set(CMAKE_${lang}_COMPILER_FORCED ON)
endforeach()
```

After this I got another error:
```
MSVC_RUNTIME_LIBRARY value 'MultiThreadedDebugDLL' not known for this OBJC compiler.
```

This can be worked around by resetting the corresponding runtime library option to an empty string (note that this must be done before the `project()` call, or before `enable_language(OBJC)`):
```cmake
foreach(lang IN ITEMS OBJC OBJCXX)
    set(CMAKE_${lang}_COMPILER_FORCED ON)

    foreach(runtimeLibrary IN ITEMS MultiThreaded MultiThreadedDLL MultiThreadedDebug MultiThreadedDebugDLL)
        set(CMAKE_${lang}_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY_${runtimeLibrary} "")
    endforeach()
endforeach()
```

While these workarounds seems to work fine, I was wondering if you might have any idea about what is missing for CMake on Windows to handle this correctly?

davidchisnall:
It might be worth opening bugs with CMake.  I suspect that they'd be willing to support more Objective-C platforms but without anyone on the CMake project using them they might not know that things are broken.

triplef:
Makes sense, I’ll do that for the Windows error.

However I’m not sure about the second issue I wrote about, is it even possible to fix this in CMake? If there’s no runtime built yet, CMake couldn’t possible verify the ObjC compiler is working I think?

CMakeLists.txt
davidchisnall: I wanted to change to just depending on the C++ standard library.  We don't gain much by avoiding that as a dependency and we lose the ability to depend on C++ standard library functions.

davidchisnall:
> However I’m not sure about the second issue I wrote about, is it even possible to fix this in CMake? If there’s no runtime built yet, CMake couldn’t possible verify the ObjC compiler is working I think?

I'm not sure.  This was the problem that I had with depending on GNUstep Make originally: it expected that Objective-C projects would have a working Objective-C stack.  For libcxxrt, I use a CMake flag that tells it to link as C, which avoids it adding the C++ libraries, but I don't know if that approach works for Objective-C.

davidchisnall:
> While these workarounds seems to work fine

I haven't been able to reproduce this, the Windows build is still failing for me.  Can you push something to this branch with your fixes?

triplef:
I’ve fixed the compiler detection on Windows. The `COMPILER_FORCED` option must be set before CMake configures the compiler, which happens as part of the `project()` call if the language is listed. So to work around this we need to remove OBJC(XX) from the `project()` call and then enable the language after forcing the compiler like so (setting these options before `project()` also didn’t work for me):

```cmake
project(libobjc C ASM CXX)

# fix up CMake Objective-C compiler detection on Windows before enabling languages below
if (WIN32)
	foreach(lang IN ITEMS C CXX)
		set(CMAKE_OBJ${lang}_COMPILER_FORCED ON)
		foreach(runtimeLibrary IN ITEMS MultiThreaded MultiThreadedDLL MultiThreadedDebug MultiThreadedDebugDLL)
			set(CMAKE_OBJ${lang}_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY_${runtimeLibrary} ${CMAKE_${lang}_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY_${runtimeLibrary}})
		endforeach()
	endforeach()
endif()

enable_language(OBJC OBJCXX)
```

However there’s still a build error, I’m guessing because we need to add `-fobjc-exceptions` to `CMAKE_OBJC(XX)_FLAGS`?
```
associate.m(171,2): error: cannot use '@try' with Objective-C exceptions disabled
```

davidchisnall:
Can you push the fix to this branch?

triplef:
Yes already done.

davidchisnall:
> However there’s still a build error, I’m guessing because we need to add -fobjc-exceptions to CMAKE_OBJC(XX)_FLAGS?

I think so, I'm a bit surprised it isn't in the default CMake flags.  We already explicitly add `-fexceptions`.

triplef:
CMake [doesn’t seem to have](https://gitlab.kitware.com/search?search=fobjc-exceptions&nav_source=navbar&project_id=541&group_id=415&search_code=true&repository_ref=fff8f3bee92d9f139c48cdc194ca826447fc402a) any reference to `-fobjc-exceptions`. Is that part of the compiler defaults on other platforms maybe?

In any case, I added it in CMakeLists, but now a bunch of tests are failing and there are these compiler warnings:
```
clang-cl: warning: unknown argument ignored in clang-cl: '-x' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-std=gnu17' [-Wunknown-argument]
clang-cl: warning: objective-c: 'linker' input unused [-Wunused-command-line-argument]
```

If I build with clang.exe instead of clang-cl.exe (locally) I don’t get these warnings and all tests pass. Any idea what else we might need to do to make clang-cl happy?

weliveindetail:
Just a note that you want to add `-fobjc-arc-exceptions` as well (if it's not added implicitly). See details here: https://reviews.llvm.org/D137939?id=475122#inline-1333144

davidchisnall:
Not all of the Objective-C things are supported by default in clang-cl, because it aims for compatibility with cl.exe (which doesn't support Objective-C).  You can pass clang arguments to the clang front end through the clang-cl driver with `-Xclang -{argument}`.

davidchisnall:
The reason that I was using clang-cl, not clang, originally was that CMake conflated using-the-gcc-like-front-end with using-a-unix-compat-layer in a bunch of places.  I think this was fixed around 3.19, so it's possible that we can just use clang.exe (which natively supports all of the Objective-C flags) instead now?

triplef:
Yes that would be the easiest solution to just switch the CI to use clang instead of clang-cl – I’ve just pushed a change for that.

I also tried to compare clang-cl invocations between the master branch and this, and one difference is that on master the source file is passed as `-c -- <source.m>`, whereas on this branch the `--` is missing. I’m not sure what this does but maybe this explains the `'linker' input unused` warning?

davidchisnall:
If we're not using clang-cl, we may not need the horrible `CCC_OVERRIDE_OPTIONS` hack.  This uses a debug feature to rewrite the arguments because CMake was adding the cl.exe flags for C or C++ to the command line in a way that could not be changed.  

The -- normally just indicates the end of flags, so won't treat files with flag-like names after it as flags.  It shouldn't make a difference here.

davidchisnall:
If you add `-v` to the ninja invocation in CI, it will show the complete command for each build step, which might help debugging.

CMakeLists.txt
davidchisnall: I think I didn't make all of those changes.

We originally linked to the C++ runtime directly for two reasons:

 - We needed to know what it was to accommodate subtle ABI differences.
 - We wanted to build without any C++ support

The first of these doesn't matter anymore - we dynamically detect the runtime.

The second is no longer something that I care about.  We use C++ features in non-optional code paths and I'd rather be able to depend on a C++ standard library existing.  As such, I'd like to rip out all of the runtime detection logic and just link as C++ to pull in a standard library implementation.

triplef: Ah I see, thanks for the explanation. I was just concerned that the logging change indicated an issue, but it sounds like it’s expected. I did just run our Android app with a new library build from this branch and it’s working as expected.

triplef (approved):
Looks good to me!

Should we still remove `CCC_OVERRIDE_OPTIONS` in main.yml and CMakeLists? Additionally we could output a warning if clang-cl is used.

davidchisnall:
> Should we still remove `CCC_OVERRIDE_OPTIONS` in main.yml and CMakeLists? Additionally we could output a warning if clang-cl is used.

Yup, I think both of those are good ideas.

triplef:
I’ve done these updates. 

Only downside I see with not supporting clang-cl is that I think it’s no longer possible to build with Visual Studio, as that doesn’t seem to support using clang according to [this](https://discourse.cmake.org/t/how-to-specify-gnu-like-clang-for-visual-studio-16-2019-generator/6639/2). Do you see an issue with that? If not I’ll update [README.windows](https://github.com/gnustep/libobjc2/blob/master/README.windows) accordingly.

triplef:
Just saw there was a test failure in the most recent run, but I don’t think it’s related to my changes:
```
88/90 Test #29: ManyManySelectors ........................Exit code 0xc0000409
***Exception:   1.59 sec
'345067selector543eb' != '345066selector543ea'
Assertion failed: strcmp(selBuffer, sel_getName(nextSel)) == 0, file D:/a/libobjc2/libobjc2/Test/ManyManySelectors.m, line 38
```

davidchisnall:
That test seems to fail reliably on some of the QEMU things.  I spent some time a couple of weeks ago trying to debug it.  I strongly suspect a bug in my hash table code, so I plan on moving the selector table over to the tsl one soon to see if that fixes it.

davidchisnall:
> Only downside I see with not supporting clang-cl is that I think it’s no longer possible to build with Visual Studio, as that doesn’t seem to support using clang according to [this](https://discourse.cmake.org/t/how-to-specify-gnu-like-clang-for-visual-studio-16-2019-generator/6639/2). Do you see an issue with that? If not I’ll update [README.windows](https://github.com/gnustep/libobjc2/blob/master/README.windows) accordingly

That's slightly unfortunate.  I did debug the tests in Visual Studio a few times, though I eventually moved to WinDbg (time travel debugging is *amazing* for the test cases: I can run them until they crash and then set a watchpoint on the location of memory corruption and rewind them until I get to the point where the corruption happened).  It sounds as if VS can drive CMake  + Ninja, so that's probably fine.

triplef:
Just FYI I believe even with these workarounds we put in place for Windows it still requires a newer CMake version than the current 3.16 minimum. I tried building with 3.20 and got the following error on Windows:

```
CMake Error at CMakeLists.txt:210 (add_library):
OBJC_STANDARD is set to invalid value ''
```

With CMake 3.24 it works, but I’m not sure which exact release fixes the issue. I just wanted to leave this here in case anyone else runs into this.

Issue #222, opened 2022-01-18
Exception handling on Windows causing access violation with ARC


triplef:
Throwing/catching an exception in ARC-enabled code on Windows causes an access violation. I’ve reproduced this in various environments/projects and both x64 and x86 with the same results. While debug builds will usually crash almost immediately, release builds can sometimes continue execution for a bit before crashing.

I also tried using Application Verifier and Page Heap to find the cause, but unfortunately I never got any more useful info. The stack trace is usually corrupt and/or unhelpful.

This can for example be reproduced with the ObjCXXEHInterop_arc test, which is [currently disabled](https://github.com/gnustep/libobjc2/commit/249acabbf4f9bd0d92f33d6a6153a06769ea91b4) for Windows due to the failure:

```
Exception thrown at 0x00007FFA9FC522C8 (objc.dll) in ObjCXXEHInterop_arc.exe: 0xE06D7363: Microsoft C++ Exception (parameters: 0x0000000019930520, 0x0000003160DFF6A8, 0x0000003160DFF5D8, 0x0000003160DFF6A7).
ObjCXXEHInterop_arc.exe has triggered a breakpoint.

ObjCXXEHInterop_arc.exe has triggered a breakpoint.

Exception thrown at 0x00007FFA9FC25280 (vcruntime140d.dll) in ObjCXXEHInterop_arc.exe: 0xC0000005: Access violation executing location 0x00007FFA9FC25280.
```

Any thoughts on how to debug this further would be appreciated.

davidchisnall:
Normally I'd debug this kind of thing in WinDbg using its time-travel mode: get to the crash, look where the pointer to the faulting address came from, and then run in reverse with a watchpoint set to that address and see where it happened.  It may be that we're failing to retain / release over exception throwing (though I thought the compiler did that in ARC mode?), but if you can find out what 0x00007FFA9FC25280 used to point to with the time-travel debugger then you can see where it was deallocated.

triplef:
That sounds like a good plan. Unfortunately I haven’t really used WinDbg before and are not really getting anywhere with it...

For starters, could you tell me how to find out where the pointer to the faulting address came from?

I used [this tutorial](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/time-travel-debugging-walkthrough) as guideline and are getting a list of events that shows the C++ exception that is thrown directly followed by the access violation exception. It would be great if you could give me some pointers where to go from there.

[...]

PR #226, opened 2022-04-27
Workaround bug in test ObjCXXEHInterop_arc

weliveindetail:
The test failure appears to origin from a Clang bug that is related to the CodeGen passes [PreISelIntrinsicLowering](https://github.com/llvm/llvm-project/blob/38e802a09de645b54c3fc983d692a1436bfc7d13/llvm/lib/CodeGen/PreISelIntrinsicLowering.cpp#L70) and [WinEHPrepare](https://github.com/llvm/llvm-project/blob/88bc24a7e39ef3f85857c3d0857e0e5c93b50bbc/llvm/lib/CodeGen/WinEHPrepare.cpp#L972). In simple cases like this test, the workaround is effective. In complex cases it can cause a crash in Clang further down the line. Would it make sense to pull it into upstream libobjc2 until we have a fix in mainline Clang and a release that contains it?

The --disable-cleanups flag used here is not as general-purpose as it may sound. It's used in `WinEHPrepare.cpp` exclusively and only triggers a single branch.

Happy to talk about an actual fix in more detail when time permits!

### DIFF
--- a/Test/CMakeLists.txt
+++ b/Test/CMakeLists.txt
@@ -35,7 +35,6 @@ set(TESTS
 	PropertyAttributeTest.m
 	ProtocolExtendedProperties.m
 	PropertyIntrospectionTest.m
-	PropertyIntrospectionTest2_arc.m
 	ProtocolCreation.m
 	ResurrectInDealloc_arc.m
 	RuntimeTest.m
--- a/Test/CMakeLists.txt
+++ b/Test/CMakeLists.txt
@@ -44,7 +44,6 @@ set(TESTS
 	WeakReferences_arc.m
 	WeakImportClass.m
 	ivar_arc.m
-	ivar_atomic.m
 	IVarOverlap.m
 	IVarSuperclassOverlap.m
 	objc_msgSend.m
--- a/Test/CMakeLists.txt
+++ b/Test/CMakeLists.txt
@@ -144,12 +144,15 @@ addtest_variants("ForwardDeclareProtocolAccess" "ForwardDeclareProtocolAccess.m;
 if (ENABLE_OBJCXX)
 	addtest_variants(ObjCXXEHInterop "ObjCXXEHInterop.mm;ObjCXXEHInterop.m" true)
 	addtest_variants(ObjCXXEHInteropTwice "ObjCXXEHInteropTwice.mm" true)
-	# This test is failing on Win32, but not for any obvious reason.  Disable
-	# it for now to keep CI happy.
-	if (WIN32)
-	else()
-		addtest_variants(ObjCXXEHInterop_arc "ObjCXXEHInterop_arc.mm;ObjCXXEHInterop_arc.m" true)
-	endif()
+	addtest_variants(ObjCXXEHInterop_arc "ObjCXXEHInterop_arc.mm;ObjCXXEHInterop_arc.m" true)
+	# Disable the cleanup step in LLVM's CodeGen pass `WinEHPrepare` in order to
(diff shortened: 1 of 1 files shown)


weliveindetail:
This aims to workaround the bug tracked in https://github.com/gnustep/libobjc2/issues/222

weliveindetail:
The bot failed before running tests in build step:
```
[112/229] Building C object Test\CMakeFiles\PropertyIntrospectionTest2_arc_optimised.dir\PropertyIntrospectionTest2_arc.m.obj
```

davidchisnall:
The build failure is because newer versions of VS install a non-functional `stdatomic.h`.  These tests should work fine if `stdatomic.h` doesn't exist.  Can disable the attempt to use `stdatomic.h` on Windows for now?

weliveindetail:
Hi David, thanks for your quick reply. Right, that seems to be the issue. In fact `PropertyIntrospectionTest2_arc` is the only test that uses `stdatomic.h` directly! I dropped the include locally and the test won't compile, because it does use `atomic_bool` in a few places.

For the sake of getting the bot to run the `ObjCXXEHInterop_arc` test at least once, do you think it's acceptable to disable `PropertyIntrospectionTest2_arc` temporarily?

I'd propose to add a commit for that and revert it afterwards. Then a squashing merge could eliminate the two right? What do you think?

davidchisnall:
Can you guard the `#include` and the use of `atomic_bool` with a Windows-specific macro so that we can skip the part of the test but keep running the rest?

weliveindetail:
[The remaining tests passed](https://dev.azure.com/gnustep/e3d06b1c-9d1f-4aef-966e-70c0f260f8e2/_apis/build/builds/509/logs/151):
```
2022-04-28T12:08:53.9584777Z 85/92 Test #87: ObjCXXEHInterop ..........................   Passed    0.05 sec
2022-04-28T12:08:53.9594245Z       Start 91: ObjCXXEHInterop_arc
2022-04-28T12:08:53.9600823Z 86/92 Test #89: ObjCXXEHInteropTwice .....................   Passed    0.04 sec
2022-04-28T12:08:53.9608378Z       Start 92: ObjCXXEHInterop_arc_optimised
2022-04-28T12:08:53.9609288Z 87/92 Test #78: hash_test_optimised ......................   Passed    2.91 sec
2022-04-28T12:08:53.9610161Z 88/92 Test #92: ObjCXXEHInterop_arc_optimised ............   Passed    0.02 sec
2022-04-28T12:08:53.9610852Z 89/92 Test #90: ObjCXXEHInteropTwice_optimised ...........   Passed    0.03 sec
2022-04-28T12:08:53.9611564Z       Start 31: ManyManySelectors
2022-04-28T12:08:53.9672148Z 90/92 Test #91: ObjCXXEHInterop_arc ......................   Passed    0.03 sec
2022-04-28T12:08:56.4648789Z 91/92 Test #31: ManyManySelectors ........................   Passed    2.50 sec
2022-04-28T12:08:56.4651123Z       Start 32: ManyManySelectors_optimised
2022-04-28T12:08:58.8975918Z 92/92 Test #32: ManyManySelectors_optimised ..............   Passed    2.43 sec
2022-04-28T12:08:58.8989639Z 
2022-04-28T12:08:58.8990467Z 100% tests passed, 0 tests failed out of 92
```
So the workaround is effective for ObjCXXEHInterop_arc, which is what this PR is about.

weliveindetail:
Hi David, sorry for the late follow-up. Actually, this PR is outdated. Instead of using the workaround, we should enable this test for Clang versions 15+. Let me put together a new PR for that.

PR #229, opened 2022-07-25
Add returns-retained annotation to class_createInstance.

davidchisnall:


objc/runtime.h
```diff
@@ -26,6 +26,12 @@ extern "C" {
 #	endif
 #endif
 
+#if __has_attribute(ns_returns_retained)
```
triplef, line 29: Should this be enclosed in `#ifndef OBJC_RETURNS_RETAINED`?

davidchisnall: I don't think so, this header shouldn't be included twice and it's not defined anywhere else.  The above macro is, because it's usually defined by clang.

objc/runtime.h
```diff
@@ -289,7 +295,7 @@ Protocol *__unsafe_unretained* class_copyProtocolList(Class cls, unsigned int *o
 /**
  * Creates an instance of this class, allocating memory using malloc.
  */
-OBJC_PUBLIC
+OBJC_PUBLIC OBJC_RETURNS_RETAINED
 id class_createInstance(Class cls, size_t extraBytes);
```
triplef, line 299: Not sure if it matters, but in Apple headers the attribute comes after the function instead of before.

triplef (commented):
Unfortunately this doesn’t seem to make a difference for me regarding https://github.com/llvm/llvm-project/issues/56056. I am also getting this warning (whether the attribute comes first or last), which is also visible in the CI builds:

```
objc/runtime.h(299,55): warning: 'ns_returns_retained' attribute only applies to functions that return an Objective-C object [-Wignored-attributes]
id class_createInstance(Class cls, size_t extraBytes) OBJC_RETURNS_RETAINED;
```

It looks like the `id` return type is not recognized by Clang somehow.

davidchisnall:
I'm a bit confused.  With this change, clang generates the right IR for me.  The error is coming from C files, where `id` is `struct objc_object*` and so isn't treated as an Objective-C type.  We should gate that attribute on `defined(__OBJC__)`.

triplef:
Sorry, you are correct. The change does fix the issue mentioned in https://github.com/llvm/llvm-project/issues/56056.

Also using `#if defined(__OBJC__) && __has_attribute(ns_returns_retained)` will fix the warnings.

I was looking at another crash with LLVM 14 when accessing GNUstep Base classes for the first time (e.g. calling `[NSUserDefaults standardUserDefaults]` crashes in `_objc_block_trampoline_end_sret`), which I incorrectly thought was the same issue. I will look into that in more detail and report it separately.

objc/runtime.h
triplef, line 29: I think the following would work to resolve the warnings when importing this from non-ObjC files:
```suggestion
#if defined(__OBJC__) && __has_attribute(ns_returns_retained)
```

PR #231, opened 2022-09-10
Don't pass ANDROID_ARM_NEON when testing C++ runtime

triplef:
Fixes building for Android arm64-v8a with NDK r25, which no longer supports this flag.

Since we were passing this unconditionally we were previously inadvertently disabling NEON support if the flag wasn't set, which is likely not what anyone would have wanted as NEON is enabled by default since NDK r21 and can no longer be disabled with r25.

Fixes https://github.com/gnustep/tools-android/issues/12.


triplef:
@davidchisnall would it be ok to merge this? It’s currently blocking the tools-android CI to succeed.

For reference, these flags were originally added in #113.

davidchisnall (approved):

PR #233, opened 2022-11-17
Enable test ObjCXXEHInterop_arc on Windows for Clang 16+

weliveindetail:
This test used to fail on Windows, because Clang's codegen and transformation haven't been in-line with the requirements of the WinEH backend in LLVM. The following changes implement the missing pieces and let the test to pass in both debug and release mode:

https://reviews.llvm.org/D128190 [WinEH] Apply funclet operand bundles to nounwind intrinsics that lower to function calls in the course of IR transforms
https://reviews.llvm.org/D134441 [ObjC][ARC] Fix target register for call expanded from CALL_RVMARKER on Windows 
https://reviews.llvm.org/D137939 [CGObjC] Open cleanup scope before SaveAndRestore CurrentFuncletPad and push CatchRetScope early
https://reviews.llvm.org/D137944 [ObjC][ARC] Teach the OptimizeSequences step of ObjCARCOpts about WinEH funclet tokens

### DIFF
--- a/Test/CMakeLists.txt
+++ b/Test/CMakeLists.txt
@@ -144,9 +144,10 @@ addtest_variants("ForwardDeclareProtocolAccess" "ForwardDeclareProtocolAccess.m;
 if (ENABLE_OBJCXX)
 	addtest_variants(ObjCXXEHInterop "ObjCXXEHInterop.mm;ObjCXXEHInterop.m" true)
 	addtest_variants(ObjCXXEHInteropTwice "ObjCXXEHInteropTwice.mm" true)
-	# This test is failing on Win32, but not for any obvious reason.  Disable
-	# it for now to keep CI happy.
-	if (WIN32)
+	if (WIN32 AND CMAKE_CXX_COMPILER_ID STREQUAL Clang)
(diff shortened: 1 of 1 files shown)


weliveindetail:
This should be ready once [D137939](https://reviews.llvm.org/D137939) and [D137944](https://reviews.llvm.org/D137944) landed

weliveindetail:
```
The following tests FAILED:
	  7 - ARCTest_arc
	  8 - ARCTest_arc_optimised
	 41 - PropertyIntrospectionTest2_arc
	 42 - PropertyIntrospectionTest2_arc_optimised
```
These tests will pass in release mode as well with Clang 16+

davidchisnall:
Can you disable those for clang <16 and then we can merge this and have a follow-on to bump the clang version in CI once it's released?

weliveindetail:
This should be ready to merge now. Thx

davidchisnall (approved):

weliveindetail:
First 16.x Release Candidates are out: https://github.com/llvm/llvm-project/releases It might be worth double-checking that my CMake changes do enable the tests on Windows with Clang 16 and that they actually succeed in release mode (and not only work for me).

triplef:
Confirmed with Clang 16 rc2:
image

PR #239, opened 2023-01-29
Attempt to add QEMU to CI.

davidchisnall:


davidchisnall:
@triplef, AArch64 is passing, but 32-bit Arm is not.  Have you seen those failures?

triplef:
This is great, thanks. I haven’t seen those errors before though. Could it be because Android [seems to use](https://developer.android.com/ndk/guides/abis#sa) ARMel instead of ARMhf?

Maybe we could add ARMel as well? I was going to add this but couldn’t figure out the name of the right name of the linker library:
```yaml
          - name: armel
            system-processor: arm
            triple: arm-linux-gnueabi
            rtld: ld-linux-armhf.so.3 # correct name?
```

davidchisnall:
I think the runners have the Android SDK installed, so maybe already have an Android-specific emulator that we can use for Android testing?

triplef:
Ah yes that would be even better. Unfortunately I never took the time to figure out how to run CMake tests on Android (I had previously only copied them by hand onto the device/emulator via `adb`).

It looks like it should be possible though with some massaging – I’ll try to take a look. I found these links that might helpful:
https://cmake.org/cmake/help/latest/module/AndroidTestUtilities.html
https://gitlab.xiph.org/xiph/opus/-/merge_requests/28/diffs

davidchisnall:
32-bit Arm C++ exception tests are 'fixed' by disabling those tests in CI.  That's not really satisfying, but at least it should prevent more regressions being introduced.  32-bit Arm has a weird exception model and I'm not sure that the C++ exception interop bits *ever* worked there, but at least we won't break anything more with this merged.

PR #248, opened 2023-10-04
Prefer system-provided robin-map

qmfrederik:
[A lot of distributions ship with robin-map in their package management system](https://repology.org/project/robin-map/versions), so it may not be a hard requirement to have robin-map as a submodule.

This PR updates the build system to search for `tsl/robin_map.h` and use that one if available; if not, fall back to the submodule.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,18 +122,13 @@ endif (WIN32)
 
 find_package(tsl-robin-map)
 
-if (tsl-robin-map_FOUND)
-	include_directories("${tsl-robin-map_INCLUDE_DIRS}")
-else ()
-	if (NOT EXISTS "${CMAKE_SOURCE_DIR}/third_party/robin-map/include/tsl/robin_map.h")
-		message(FATAL_ERROR "Git submodules not present, please run:\n\n"
-							" $ git submodule init && git submodule update\n\n"
-							"If you did not checkout via git, you will need to"
-							"fetch the submodule's contents from"
-							"https://github.com/Tessil/robin-map/")
-	endif ()
+if (NOT tls-robin-map_FOUND)
+	FetchContent_Declare(
+		robinmap
+		GIT_REPOSITORY https://github.com/Tessil/robin-map/
+		GIT_TAG        7d37a41003bfbc7e12e34601f93c18ca2ff6d7c07)
 
-	include_directories("${CMAKE_SOURCE_DIR}/third_party/robin-map/include/")
+	FetchContent_MakeAvailable(robinmap)
 endif()
 
 if (WIN32)
@@ -242,6 +237,7 @@ if (WIN32)
 	target_link_libraries(objc ntdll.dll)
 endif()
 
+target_link_libraries(objc tsl::robin_map)
 
 set_target_properties(objc PROPERTIES
 	LINKER_LANGUAGE C
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,14 +120,19 @@ else ()
 	list(APPEND libobjc_C_SRCS eh_personality.c)
 endif (WIN32)
 
-if (NOT EXISTS "${CMAKE_SOURCE_DIR}/third_party/robin-map/include/tsl/robin_map.h")
-	message(FATAL_ERROR "Git submodules not present, please run:\n\n"
-						" $ git submodule init && git submodule update\n\n"
-						"If you did not checkout via git, you will need to"
-						"fetch the submodule's contents from"
-						"https://github.com/Tessil/robin-map/")
-endif ()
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -120,14 +120,19 @@ else ()
...
 endif (WIN32)
 
-if (NOT EXISTS "${CMAKE_SOURCE_DIR}/third_party/robin-map/include/tsl/robin_map.h")
-	message(FATAL_ERROR "Git submodules not present, please run:\n\n"
-						" $ git submodule init && git submodule update\n\n"
-						"If you did not checkout via git, you will need to"
-						"fetch the submodule's contents from"
-						"https://github.com/Tessil/robin-map/")
-endif ()
+find_path(ROBIN_MAP_HEADER NAMES "tsl/robin_map.h")
```
2xsaiko, line 123:
```suggestion
find_package(tsl-robin-map)
```

It should come with a CMake package. You'll also need to do something with the found file, right? I think ROBIN_MAP_HEADER is unused right now except for checking if it's there.

```cmake
target_link_libraries(objc PRIVATE tsl::robin_map)
```

CMakeLists.txt
```diff
@@ -120,14 +120,19 @@ else ()
...
+if (NOT ROBIN_MAP_HEADER)
+	if (NOT EXISTS "${CMAKE_SOURCE_DIR}/third_party/robin-map/include/tsl/robin_map.h")
+		message(FATAL_ERROR "Git submodules not present, please run:\n\n"
+							" $ git submodule init && git submodule update\n\n"
+							"If you did not checkout via git, you will need to"
+							"fetch the submodule's contents from"
+							"https://github.com/Tessil/robin-map/")
+	endif ()
 
+	include_directories("${CMAKE_SOURCE_DIR}/third_party/robin-map/include/")
```
2xsaiko, line 134:
```suggestion
	add_library(tsl::robin_map INTERFACE IMPORTED)
	set_target_properties(tsl::robin_map PROPERTIES
	    INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_SOURCE_DIR}/third_party/robin-map/include/")
```

and then it can be linked as above

2xsaiko (changes requested):
A bit more modern CMake style

davidchisnall:
It would be good to have this in the next release.  Do either of you want to incorporate the changes and update the PR?

CMakeLists.txt
qmfrederik: Correct, that works because the headers are installed into `/include/`.  I've added:
```cmake
include_directories("${tsl-robin-map_INCLUDE_DIRS}")
```
but it's redundant in most cases.

The package is a header-only package; there is no library to link against.

qmfrederik:
I've addressed the feedback (with some comments) and rebased on master.  Let me know if there's anything else you need.  Thanks!

CMakeLists.txt
davidchisnall: CMake models header-only libraries just like other libraries. If you import the target then it adds the relevant flags to find the headers. It also adds any compile flags that the ‘library’ needs.

CMakeLists.txt
```diff
@@ -105,14 +105,19 @@ else ()
...
-	message(FATAL_ERROR "Git submodules not present, please run:\n\n"
-						" $ git submodule init && git submodule update\n\n"
-						"If you did not checkout via git, you will need to"
-						"fetch the submodule's contents from"
-						"https://github.com/Tessil/robin-map/")
-endif ()
+find_path(ROBIN_MAP_HEADER NAMES "tsl/robin_map.h")
+
+if (NOT ROBIN_MAP_HEADER)
+	if (NOT EXISTS "${CMAKE_SOURCE_DIR}/third_party/robin-map/include/tsl/robin_map.h")
```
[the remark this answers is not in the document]
davidchisnall: If it isn’t found, we should use FetchContent to get it and then add it explicitly, then fall through to add_library below.

selector_table.cc
```diff
@@ -11,7 +11,7 @@
 #include <vector>
 #include <mutex>
 #include <forward_list>
-#include "third_party/robin-map/include/tsl/robin_set.h"
+#include "tsl/robin_set.h"
```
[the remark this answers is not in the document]
davidchisnall: This needs to be changed, the header 8s now found via -I. If you turn up warnings a bit, clang will tell you that this is wrong.

qmfrederik:
@davidchisnall Sorry for the noise.  This should be good now.  Can you kick off the GitHub CI for this PR, so we get the build signal on all systems?

2xsaiko (approved):

davidchisnall (approved):

davidchisnall:
Thanks, looks great.

Issue #247, opened 2023-10-04
UnexpectedException test failing on ARM ABIs


triplef:
This test was recently added in #220 and subsequently disabled on ARM (https://github.com/gnustep/libobjc2/commit/226455bd10d8b4a0a090449b5b7ebf1e42daf0a5) due to failing on the cross-build ARM CI targets.

I’m able to reproduce it on an arm64 Android emulator with this result in gdb:
```
Program received signal SIGABRT, Aborted.
0x0000007fbdee5f74 in abort () from target:/apex/com.android.runtime/lib64/bionic/libc.so
(gdb) bt
#0  0x0000007fbdee5f74 in abort () from target:/apex/com.android.runtime/lib64/bionic/libc.so
#1  0x0000007fbe2e7178 in objc_exception_throw (object=0x5555558ee0 <.objc_str_Exception>)
    at eh_personality.c:261
#2  0x0000005555556a4c in main ()
    at UnexpectedException.m:38
```

Any idea what might be going on here? Interestingly the exception hook is working fine in our app on Android ARM devices and has been for years.

Steps to reproduce with an Android emulator (on macOS ARM host, NDK paths may vary):
```
adb push libobjc.so /data/local/tmp
adb push Test/UnexpectedException /data/local/tmp
adb push $ANDROID_NDK_ROOT/sources/cxx-stl/llvm-libc++/libs/arm64-v8a/libc++_shared.so /data/local/tmp

# run test
adb shell LD_LIBRARY_PATH=/data/local/tmp/ /data/local/tmp/UnexpectedException

# debug with gdb
adb push $ANDROID_NDK_ROOT/prebuilt/android-arm64/gdbserver/gdbserver /data/local/tmp
adb forward tcp:5039 tcp:5039
adb shell
> cd /data/local/tmp
> LD_LIBRARY_PATH=$PWD ./gdbserver :5039 ./UnexpectedException

# on host machine
$ANDROID_NDK_ROOT/prebuilt/darwin-x86_64/bin/gdb
> target remote :5039
> c
```

davidchisnall:
This abort is at the end of the throw function. The unwind library is returning from the unwind function, which happens only if unwinding fails. For some reason, it looks as if it is not reporting the end of the stack as the reason. Can you see what the value of err is? If you define DEBUG_EXCEPTIONS At the top of the file, it will log it (and a load of other info) for you.

triplef:
These are the logs I get:

```
Exception caught by C++: 0
[...]

PR #249, opened 2023-11-25
Initial Windows on ARM (AArch64) Support

hmelder:
This pull request provides initial support for Windows on ARM.

- Use [FlushInstructionCache](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-flushinstructioncache) instead of unsupported clear_cache intrinsic
(https://github.com/llvm/llvm-project/issues/58416)
- Add `ntdll.dll` to `target_link_libraries` as linker is unable to find `RtlRaiseException`
- Export symbols in `objc_msgSend.aarch64.S` (PE/COFF)
- Structured Exception Handling in `objc_msgSend.aarch64.S`

The code was tested on the following Systems:

1. Windows 10 Version 22598 MP (5 procs) Free ARM 64-bit (AArch64)
Edition build lab: 22598.1.arm64fre.ni_release.220408-1503
2. Debian GNU/Linux trixie/sid aarch64 (Kernel: 6.5.0-3-arm64)

I have documented the use of the SEH directives at the beginning of the file.

Further reading material (sorted by relevance):
- https://learn.microsoft.com/en-us/cpp/build/arm64-exception-handling?view=msvc-170
- https://github.com/llvm/llvm-project/commit/cbd8464595220b5ea76c70ac9965d84970c4b712
- https://www-user.tu-chemnitz.de/~heha/hsn/chm/Win32SEH.chm/Win32SEH.htm

Thank you @mstorsjo for the detailed explaination of the .seh* directives and the link to the LLVM commit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -353,7 +353,12 @@ endif()
 # threading implementation (we do for everything except thread-local storage)
 set(CMAKE_THREAD_PREFER_PTHREAD)
 include(FindThreads)
+# Link against ntdll.dll for RtlRaiseException
+if (WIN32)
+target_link_libraries(objc Threads::Threads ntdll.dll)
+else()
 target_link_libraries(objc Threads::Threads)
+endif()
 
 
 
--- a/block_to_imp.c
+++ b/block_to_imp.c
@@ -177,7 +177,12 @@ static struct trampoline_set *alloc_trampolines(char *start, char *end)
 	}
 	metadata->buffers->headers[HEADERS_PER_PAGE-1].block = NULL;
 	mprotect(metadata->buffers->rx_buffer, PAGE_SIZE, PROT_READ | PROT_EXEC);
+#if defined(_WIN32) && (defined(__arm__) || defined(__aarch64__))
+  FlushInstructionCache(GetCurrentProcess(), start, end - start);
+#else
 	clear_cache(metadata->buffers->rx_buffer, &metadata->buffers->rx_buffer[PAGE_SIZE]);
+#endif
+
 	return metadata;
 }
 
--- a/objc_msgSend.aarch64.S
+++ b/objc_msgSend.aarch64.S
@@ -1,5 +1,16 @@
 #define ARGUMENT_SPILL_SIZE (8*10 + 8*16)
-.macro MSGSEND receiver, sel
+
+#ifdef _WIN64
+#	define START_PROC(x) .seh_proc x
+#	define END_PROC(x) .seh_endproc
+#	define FRAME_OFFSET(x) .seh_stackalloc x
+#else
+#	define START_PROC(x) .cfi_startproc
+#	define END_PROC(x) .cfi_endproc
+#	define FRAME_OFFSET(x) .cfi_adjust_cfa_offset x
+#endif
+
+.macro MSGSEND fnname, receiver, sel
 	.cfi_startproc
 	cbz    \receiver, 4f                   // Skip everything if the receiver is nil
 	                                       // Jump to 6: if this is a small object
@@ -7,6 +18,8 @@
 	cbnz   x9, 6f
 
 	ldr    x9, [\receiver]                 // Load class to x9 if not a small int
+
+.align 2
 1:
 	ldr    x9, [x9, #DTABLE_OFFSET]        // Dtable -> x9
 	ldr    w10, [\sel]                     // selector->index -> x10
@@ -35,6 +48,7 @@
 	ldr    x9, [x9, #SLOT_OFFSET]          // Load the method from the slot
 	br     x9                              // Tail-call the method
 
+.align 2
 4:	                                       // Nil receiver
 	mov    x0, #0
 	mov    v0.d[0], x0
@@ -79,21 +93,45 @@
 	ldp    \receiver, x8, [sp], #(ARGUMENT_SPILL_SIZE + 16)
 	br     x9
 6:
+#if __ELF__
 	adrp   x10, :got:SmallObjectClasses
 	ldr    x10, [x10, :got_lo12:SmallObjectClasses]
 	ldr    x9, [x10, x9, lsl #3]
+#else
+  adr    x9, CDECL(SmallObjectClasses)
+#endif
 	b      1b
 	.cfi_endproc
 .endm
 
+#ifdef _WIN64
+.text
+.def objc_msgSend;
+.scl 2;
+.type 32;
+.endef
+.def objc_msgSend_fpret;
+.scl 2;
+.type 32;
+.endef
+.def objc_msgSend_stret;
+.scl 2;
+.type 32;
+.endef
+#endif
 .globl CDECL(objc_msgSend_fpret)
 TYPE_DIRECTIVE(CDECL(objc_msgSend_fpret), %function)
 .globl CDECL(objc_msgSend)
 TYPE_DIRECTIVE(CDECL(objc_msgSend), %function)
 .globl CDECL(objc_msgSend_stret)
 TYPE_DIRECTIVE(CDECL(objc_msgSend_stret), %function)
-CDECL(objc_msgSend):
 CDECL(objc_msgSend_fpret):
 CDECL(objc_msgSend_stret):
-	MSGSEND x0, x1
-
+CDECL(objc_msgSend):
+	MSGSEND objc_msgSend x0, x1
+#ifdef _WIN64
+.section        .drectve,"yn"
+.ascii  " /EXPORT:objc_msgSend"
+.ascii  " /EXPORT:objc_msgSend_fpret"
+.ascii  " /EXPORT:objc_msgSend_stret"
+#endif
--- a/objc_msgSend.aarch64.S
+++ b/objc_msgSend.aarch64.S
@@ -61,6 +61,8 @@
 										  // that we adjust the sp before storing the pair
 										  // of registers.
 	stp    x0, x1, [sp, #-(ARGUMENT_SPILL_SIZE)]!
+	.seh_stackalloc (ARGUMENT_SPILL_SIZE)
+
 	stp    x2, x3, [sp, #16]              // The order is arbitrary, except that
 	stp    x4, x5, [sp, #32]              // fp and lr must be spilled together and
 	stp    x6, x7, [sp, #48]              // it's convenient if \receiver is spilled at sp
@@ -69,13 +71,19 @@
 	stp    q4, q5, [sp, #128]
 	stp    q6, q7, [sp, #160]
 	stp    fp, lr, [sp, #192]
+	.seh_save_fplr	192
+
 	add    fp, sp, 192					  // Adjust frame pointer
+	.seh_add_fp	192
 	stp    \receiver, x8, [sp, #-16]!     // Pre-index by sp-16
+	.seh_stackalloc 16
 
 	// FIXME: Conditionally use SEH directives
+	/*
 	.cfi_def_cfa fp, 16
 	.cfi_offset fp, -16
 	.cfi_offset lr, -8
+	*/
 	
 	                                      // We now have all argument registers, the link
 	                                      // register and the receiver spilled on the
@@ -96,7 +104,12 @@
 	ldp    q4, q5, [sp, #144]
 	ldp    q6, q7, [sp, #176]
 	ldp    fp, lr, [sp, #208]
+	.seh_save_fplr 208
+
+										  // Post-increment sp += ARGUMENT_SPILL_SIZE +16
 	ldp    \receiver, x8, [sp], #(ARGUMENT_SPILL_SIZE + 16)
+	.seh_stackalloc  (ARGUMENT_SPILL_SIZE + 16)
+
 	br     x9
 6:
 										  // Load 63:12 of SmallObjectClasses address
--- a/objc_msgSend.aarch64.S
+++ b/objc_msgSend.aarch64.S
@@ -1,18 +1,52 @@
 #define ARGUMENT_SPILL_SIZE (8*10 + 8*16)
 
+/* Windows ARM64 Exception Handling
+ *
+ * Structured Exception Handling (SEH) on Windows ARM64 differs from the x64
+ * implementation. Functions consist of a single prologue and zero or more
+ * epilogues. Instead of using offsets for the .seh* directives to manipulate the
+ * stack frame, each directive corresponds to a single instruction.
+ *
+ * This presents a challenge for our objc_msgSend function, which only modifies
+ * the stack when a slow lookup is needed (see label "5").
+ *
+ * To address this, we move the directive marking the start of a function deep
+ * into the msgSend body to prevent marking every instruction as ".seh_nop."
+ *
+ *   For 64-bit Windows (_WIN64):
+ *  - EH_START(x): Start of function (no effect on Windows)
+ *  - EH_END(x): End of function (no effect on Windows)
+ *  - EH_START_AT_OFFSET(x): Mark Start of function (Delayed)
+ *  - EH_END_AT_OFFSET(x): Mark End of function (Delayed)
+ *  - EH_END_PROLOGUE: End of function prologue
+ *  - EH_START_EPILOGUE: Start of function epilogue
+ *  - EH_END_EPILOGUE: End of function epilogue
+ *  - EH_SAVE_FP_LR(x): Save Frame Pointer and Link Register
+ *  - EH_STACK_ALLOC(x): Stack allocation (inside prologue)
+ *  - EH_ADD_FP(x): Add to Frame Pointer
+ *  - EH_NOP: Mark instruction with no unwinding relevance
+ *
+ * For non-64-bit Windows systems or other platforms, these macros have no effect and can be used without causing issues.
+ */
+
 #ifdef _WIN64
 #   define EH_START(x)
 #   define EH_END(x)
+
 #   define EH_START_AT_OFFSET(x) .seh_proc x
 #   define EH_END_AT_OFFSET(x) .seh_endproc x
+
 #   define EH_END_PROLOGUE .seh_endprologue
 #   define EH_START_EPILOGUE .seh_startepilogue
 #   define EH_END_EPILOGUE .seh_endepilogue
+
 #   define EH_SAVE_FP_LR(x) .seh_save_fplr x
 #   define EH_STACK_ALLOC(x) .seh_stackalloc x
 #   define EH_ADD_FP(x) .seh_add_fp	x
+
 #   define EH_NOP .seh_nop
 #else
+// Marks the real start and end of the function
 #   define EH_START(x) .cfi_startproc
 #   define EH_END(x) .cfi_endproc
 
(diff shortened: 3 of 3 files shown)


hmelder:
I need to rebase the branch.

CMake reports the wrong CPU architecture as we run it as an Intel process:

```sh
-- Selecting Windows SDK version 10.0.22621.0 to target Windows 10.0.22598.
-- The C compiler identification is MSVC 19.33.31517.0
-- The CXX compiler identification is MSVC 19.33.31517.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Preview/VC/Tools/MSVC/14.33.31517/bin/HostARM64/ARM64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Preview/VC/Tools/MSVC/14.33.31517/bin/HostARM64/ARM64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMAKE_SYSTEM_PROCESSOR: AMD64
-- Configuring done (3.4s)
-- Generating done (0.0s)
-- Build files have been written to: C:/Users/vm/test/build
```

The check https://github.com/gnustep/libobjc2/commit/226455bd10d8b4a0a090449b5b7ebf1e42daf0a5 fails

hmelder:
See https://github.com/gnustep/libobjc2/issues/247

CMakeLists.txt
```diff
@@ -214,7 +229,12 @@ list(APPEND libobjc_CXX_SRCS ${libobjcxx_CXX_SRCS})
 target_sources(objc PRIVATE ${libobjc_CXX_SRCS})
 
 include(FindThreads)
+# Link against ntdll.dll for RtlRaiseException
+if (WIN32)
+target_link_libraries(objc Threads::Threads ntdll.dll)
+else()
 target_link_libraries(objc Threads::Threads)
+endif()
```
davidchisnall, line 237:
```suggestion
target_link_libraries(objc Threads::Threads)
# Link against ntdll.dll for RtlRaiseException
if (WIN32)
	target_link_libraries(objc ntdll.dll)
endif()
```

CMakeLists.txt
```diff
@@ -34,6 +34,21 @@ if (MSVC)
 	set(objc_LINK_FLAGS "/DEBUG /INCREMENTAL:NO ${objc_LINK_FLAGS}")
 endif()
 
+# Get Architecture without relying on CMake
```
davidchisnall, line 37: Please can you add a comment explaining why we can't rely on CMake here?

block_to_imp.c
```diff
@@ -177,7 +177,12 @@ static struct trampoline_set *alloc_trampolines(char *start, char *end)
 	}
 	metadata->buffers->headers[HEADERS_PER_PAGE-1].block = NULL;
 	mprotect(metadata->buffers->rx_buffer, PAGE_SIZE, PROT_READ | PROT_EXEC);
+#if defined(_WIN32) && (defined(__arm__) || defined(__aarch64__))
```
davidchisnall, line 180: Please can you implement the [`clear_cache`](https://github.com/gnustep/libobjc2/blob/226455bd10d8b4a0a090449b5b7ebf1e42daf0a5/block_to_imp.c#L25) macro (which is already target-specific) with the Windows version, rather than polluting generic code with target-specific ifdefs?

objc_msgSend.aarch64.S
```diff
@@ -1,12 +1,78 @@
...
+
+#   define EH_SAVE_FP_LR(x) .seh_save_fplr x
+#   define EH_STACK_ALLOC(x) .seh_stackalloc x
+#   define EH_ADD_FP(x) .seh_add_fp	x
+
+#   define EH_NOP .seh_nop
+#else
+// Marks the real start and end of the function
+#   define EH_START(x) .cfi_startproc
+#   define EH_END(x) .cfi_endproc
```
davidchisnall, line 51: Neither of the definitions of these seem to use the argument.  Why is it there?

objc_msgSend.aarch64.S
```diff
@@ -1,12 +1,78 @@
...
+.macro MSGSEND fnname, receiver, sel
+	EH_START(\fnname)
+
 	cbz    \receiver, 4f                   // Skip everything if the receiver is nil
 	                                       // Jump to 6: if this is a small object
 	ubfx    x9, \receiver, #0, #SMALLOBJ_BITS
 	cbnz   x9, 6f
 
 	ldr    x9, [\receiver]                 // Load class to x9 if not a small int
+
```
davidchisnall, line 75:
```suggestion
```

objc_msgSend.aarch64.S
```diff
@@ -41,59 +107,128 @@
...
 TYPE_DIRECTIVE(CDECL(objc_msgSend_fpret), %function)
 .globl CDECL(objc_msgSend)
 TYPE_DIRECTIVE(CDECL(objc_msgSend), %function)
 .globl CDECL(objc_msgSend_stret)
 TYPE_DIRECTIVE(CDECL(objc_msgSend_stret), %function)
-CDECL(objc_msgSend):
 CDECL(objc_msgSend_fpret):
 CDECL(objc_msgSend_stret):
-	MSGSEND x0, x1
+CDECL(objc_msgSend):
```
davidchisnall, line 226: Why are these reordered?

objc_msgSend.aarch64.S
```diff
@@ -41,59 +107,128 @@
...
+	MSGSEND objc_msgSend x0, x1
```
davidchisnall, line 227: It would probably simplify the code to remove the macro here.  On x86, we need a bunch of different variants, but on AArch64 they're all the same.

CMakeLists.txt
hmelder: When using the x64 variant of CMake on Windows (which is the version shipped by Chocolatey), CMake reports that the CPU architecture is AMD64, and not ARM64 as expected.

When using the Preprocessor, we can reliably check the CPU architecture.

objc_msgSend.aarch64.S
hmelder: You are correct. A similar macro is in the x86_64 implementation (END_PROC) which is not using the arg either. I will fix that!

objc_msgSend.aarch64.S
hmelder: Makes debugging easier too!

CMakeLists.txt
mstorsjo: If you're getting `AMD64` here when cross compiling, I would expect that you haven't set up your cmake configuration entirely correctly for cross compilation; when cross compiling with cmake you need to set `CMAKE_SYSTEM_NAME`, even if cross compiling from Windows x86_64 to Windows aarch64 - see https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html and https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_NAME.html.

You didn't say which variable you did check, that gave you `AMD64` - I would expect `CMAKE_SYSTEM_PROCESSOR`. This one is tricky, because when you're not cross compiling, then this is set to a rather inconsistent set of values depending on platform (e.g. amd64/x86_64/x64 and arm64 vs aarch64 and similar, differ depending on OS), see https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html#variable:CMAKE_HOST_SYSTEM_PROCESSOR. When cross compiling, then CMake doesn't set `CMAKE_SYSTEM_PROCESSOR` at all, but the user configuring the cross compile is supposed to set it. As most things build fine without setting it, many cmake cross compilation setups don't set it at all.

There's another variable, `CMAKE_C_COMPILER_ARCHITECTURE_ID`, which CMake sets automatically based on detecting things from the compiler. However this is set only in some cases so one can't generally depend on it either, see https://gitlab.kitware.com/cmake/cmake/-/issues/17702 and https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6432.

So TL;DR if you want something that automatically detects this from the compiler rather than having the user specify it, then something like the suggested example probably is quite reasonable.

objc_msgSend.aarch64.S
```diff
@@ -1,12 +1,78 @@
...
+ * epilogues. Instead of using offsets for the .seh* directives to manipulate the
+ * stack frame, each directive corresponds to a single instruction.
+ *
+ * This presents a challenge for our objc_msgSend function, which only modifies
+ * the stack when a slow lookup is needed (see label "5").
+ *
+ * To address this, we move the directive marking the start of a function deep
+ * into the msgSend body to prevent marking every instruction as ".seh_nop."
+ *
+ *   For 64-bit Windows (_WIN64):
```
mstorsjo, line 16: This is a nitpick - but...

I generally very seldom see the use for checking the `_WIN64` define. If you want to check for platform/OS, then `_WIN32` is more of the canonical way. `_WIN64` was commonly used to distinguish x86_64 from i386 in Windows specific code, but when involving arm and aarch64 as well (and then it's better to explicitly check for individual architectures), it's often kinda pointless unless you explicitly want to check for the pointer size.

As this file seems to be aarch64 specific, we know the architecture already, so the OS is the only variable, so I'd just check for `_WIN32`.

Then secondly, regarding the use fo "64 bit Windows" in the text of the comment here - most of the logic for unwind info doesn't really generalize across 64 bit Windows, and where it does generalize, it's not about 32 vs 64 bit.

On i386, this kind of table based SEH unwind info isn't used at all; if building with mingw toolchains that use DWARF, then they use regular `.cfi_*` directives though.

On x86_64, `.seh_*` directives similar to on aarch64 are used - although the similarities are very few. Common ones such as `.seh_proc`, `.seh_endproc` and `.seh_stackalloc` are shared, but x86_64 doesn't have the same logic for epilogues IIRC, and all the individual directives are wildly different anyway - so I don't see much value to group them together.

On 32 bit ARM, a similar mechanism is used as on x86_64 and AArch64. The style/logic (1:1 mapping between instructions and unwind opcodes, same style with prologues/epilogues, etc) is exactly the same as for AArch64, but most of the actual opcodes differ here as well. In practice, there's little reason to care about modern Windows on 32 bit ARM (unless you care about Windows Phone, WinRT or Windows 10 IoT) though. But it's another example case against grouping unwind info style into "64 bit Windows".

I.e., TL;DR, I'd skip the wording about "64 bit Windows" and just talk about Windows in general, unless these macros actually are used in a header shared across multiple architectures.

CMakeLists.txt
hmelder: I forgot to add the patch for Tests/CMakeLists.txt where the ARCHITECTURE variable is actually used.

objc_msgSend.aarch64.S
hmelder: Thank you for the in depth answer :)

objc_msgSend.aarch64.S
hmelder: But what about the function name that changes depending on which objc_msgSend is invoked? We need the name for .seh_startproc.

triplef (approved):
Lgtm, though I can’t really comment on the assembly changes. Thank you for putting this together!

Unfortunately GitHub Actions so far doesn’t support WoA runners (see https://github.com/actions/runner-images/issues/768), but would it be possible to at least cross-compile for ARM on the current CI (without running the tests)?

objc_msgSend.aarch64.S
mstorsjo: FWIW, I just checked, and it looks like symbol name passed to `.seh_startproc` actually isn't used for anything of value. It looks like you actually could pass any random string as symbol name to it. (WIthin LLVM, it seems like the symbol you pass only is used for printing error messages.)

Also, it's rather relevant here, that when you have a `.seh_proc` that start at a different position than the function itself, the unwind info only starts covering things from the .seh_proc` location. This also means that the 1:1 instruction<->opcode correspondence works out correctly as intended.

davidchisnall: On all platforms, this function plays somewhat fast and loose with EH mechanisms.  The only thing that *should* be able to cause / propagate exceptions is the call to the slow path, which may end up throwing exceptions out of `+initialize`.  If the lookup succeeds, we tail call the real IMP and so exceptions from that message don't need to be propagated.  In both cases, we don't do any cleanup here and so just need the unwind to work correctly.

As a result, we need the correct unwind table info to be set up on that one branch instruction, we don't care if it's broken everywhere else in the function.

On 32-bit i386 Windows, the setjmp-style mechanism worked very well for us because it unwinds through any function that doesn't directly interact with SEH without running any cleanups.  Everywhere else, we do the absolute minimum to replicate this behaviour.

CMakeLists.txt
davidchisnall: This, indeed, sounds like a problem with your cross-compile setup.  If CMake is setting the wrong architecture type then all manner of other things will be broken.

hmelder: I am not cross-compiling in the traditional sense. I can either use a AMD64 or ARM64 version of CMake on WoA. Most distributions only provide a AMD64 release. The Build System is not a native process, which is fine.

Despite the difference, CMake selects the native version of LLVM (AArch64).

As @mstorsjo said:
> So TL;DR if you want something that automatically detects this from the compiler rather than having the user specify it, then something like the suggested example probably is quite reasonable.

It is hard to detect a non-native CMake without my patch. Force users to use the ARM64 version, and tolerate weird behaviour when running a non-native version?

What is your solution?

objc_msgSend.aarch64.S
hmelder: > FWIW, I just checked, and it looks like symbol name passed to `.seh_startproc` actually isn't used for anything of value. It looks like you actually could pass any random string as symbol name to it. (WIthin LLVM, it seems like the symbol you pass only is used for printing error messages.)
> 
> Also, it's rather relevant here, that when you have a `.seh_proc` that start at a different position than the function itself, the unwind info only starts covering things from the .seh_proc` location. This also means that the 1:1 instruction<->opcode correspondence works out correctly as intended.

Alright. I will just pass `objc_msgSend` then, and remove the macro.

Mushroom:
How are tests run? An option is docker+qemu+wine on an ubuntu x64 machine, such as here in sse2neon: https://github.com/DLTcollab/sse2neon/pull/598

Might not work in all cases, but possibly worth a try

anthony-linaro:
Repost from correct account :)

How are tests run? An option is docker+qemu+wine on an ubuntu x64 machine, such as here in sse2neon: https://github.com/DLTcollab/sse2neon/pull/598

Might not work in all cases, but possibly worth a try

hmelder:
> Repost from correct account :)
> 
> How are tests run? An option is docker+qemu+wine on an ubuntu x64 machine, such as here in sse2neon: [DLTcollab/sse2neon#598](https://github.com/DLTcollab/sse2neon/pull/598)
> 
> Might not work in all cases, but possibly worth a try

That would be an interesting approach! Self-hosting is also an option but requires someone to pay for an arm64 VM :/

CMakeLists.txt
davidchisnall: Can you report this upstream to CMake?  If it's finding the AArch64 tools, it should know that the target is AArch64.  I presume that there's a similar issue using an x86-64 binary under Rosetta 2 on macOS when targeting the native platform.

objc_msgSend.aarch64.S
```diff
@@ -41,59 +114,114 @@
...
-CDECL(objc_msgSend):
-CDECL(objc_msgSend_fpret):
-CDECL(objc_msgSend_stret):
-	MSGSEND x0, x1
+#ifdef _WIN32
+.text
+.def objc_msgSend;
+.scl 2;
+.type 32;
+.endef
```
davidchisnall, line 213: Not essential for this PR, but it would be nice if these were wrapped in the same sort of macros as the ELF versions (maybe replace those with explicit `FUNCTION_DEFINE` or similar that captures what we need for both)?

davidchisnall (approved):
I still don't really like the amount of hackery that we have to work around a CMake bug, but we've merged hackery to work around CMake before.  Can you open an issue that tracks the upstream CMake issue so that we can remove it once CMake is fixed?

hmelder:
We could try the [CMAKE_HOST_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html#variable:CMAKE_HOST_SYSTEM_PROCESSOR) variable as well but this will cause problems when cross compiling.

I found this in the [CMAKE_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html) documentation which is related to our problem:
> In many cases, this will correspond to the target architecture for the build, but this is not guaranteed. (E.g. on Windows, the host may be AMD64 even when using a MSVC cl compiler with a 32-bit target.)

I will open an issue, but I have low expectations.

hmelder:
It comes down on how CMake retrieves the architecture from the compiler.

anthony-linaro:
Repost from correct account (again, sorry all)

You guys are running inside a vcvarsall window, right? In blender, where I his something similar, I just checked "$ENV{VSCMD_ARG_TGT_ARCH}"

hmelder:
> Repost from correct account (again, sorry all)
> 
> You guys are running inside a vcvarsall window, right? 

Yes:
```sh
vm@localhost's password:
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.3.0-pre.2.0
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'arm64'
```

mstorsjo:
> It comes down to how CMake retrieves the architecture from the compiler.

For `CMAKE_HOST_SYSTEM_PROCESSOR`, cmake pretty much explicitly doesn’t check the compiler, only the host environment. And `CMAKE_SYSTEM_PROCESSOR` is a plain copy of the former, if not cross compiling.

For detecting things from the actual compiler, there’s only `CMAKE_C_COMPILER_ARCHITECTURE_ID`, which only is available in some cases.

anthony-linaro:
If it's always going to be in a `vcvarsall` window on Windows, I would suggest `"$ENV{VSCMD_ARG_TGT_ARCH}"` then - avoids CMake completely

hmelder:
> If it's always going to be in a `vcvarsall` window on Windows, I would suggest `"$ENV{VSCMD_ARG_TGT_ARCH}"` then - avoids CMake completely

So we need to error out, when we do not detect a vcvarsall enviroment, or fall back to some other value. Also not an ideal solution.

It is ridiculous how such an elementary thing is just completely broken in CMake.

triplef:
> So we need to error out, when we do not detect a vcvarsall enviroment

That seems ok to me since I think one always needs a Visual Studio environment on Windows, right?

anthony-linaro:
Yes, I think that's a reasonable assumption - I haven't come across anything that doesn't use it (although I'm happy to be proved wrong) - even stuff that doesn't use CL

mstorsjo:
> > So we need to error out, when we do not detect a vcvarsall enviroment
> 
> That seems ok to me since I think one always needs a Visual Studio environment on Windows, right?

If using mingw based tools (either GNU tools, or llvm tools with mingw headers), then one does not use a visual studio environment at all.

triplef:
> If using mingw based tools (either GNU tools, or llvm tools with mingw headers), then one does not use a visual studio environment at all.

Good point, but libobjc2 doesn’t currently support MinGW.

davidchisnall:
@triplef, I've made you admin.  I think that's sufficient to add runners.

CMakeLists.txt
hmelder: https://gitlab.kitware.com/cmake/cmake/-/issues/25466

objc_msgSend.aarch64.S
hmelder: I will do this in a separate PR, as the other versions of msgSend (x86_64 and x86-32) do the same thing. We can probably streamline the function definitions in Common.S.

PR #253, opened 2023-12-29
Tests: Append CMAKE_BINARY_DIR to PATH on Windows

qmfrederik:
Avoids copying objc.dll to the tests directory.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -174,7 +174,6 @@ jobs:
       shell: cmd
       working-directory: build
       run: |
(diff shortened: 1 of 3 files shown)


davidchisnall (approved):

davidchisnall:
Looks fine, but can you check that this works in a fresh terminal session? There were some exciting problems with CMake not preserving environment variable changes.

qmfrederik:
On Windows, yes, this works (I tried locally and CI also excercises that path).

PR #255, opened 2023-12-31
Don't set -fobjc-runtime=gnustep-2.0 as a global flag

qmfrederik:
Calling:

```cmake
add_compile_options("$<$<OR:$<COMPILE_LANGUAGE:OBJC>,$<COMPILE_LANGUAGE:OBJCXX>>:-Wno-deprecated-objc-isa-usage;-Wno-objc-root-class;-fobjc-runtime=gnustep-2.0>$<$<COMPILE_LANGUAGE:C>:-Xclang;-fexceptions>")
```

in `CMakeLists.txt` will add `-fobjc-runtime=gnustep-2.0` as a compile option for all builds.  The legacy tests will attempt to target `gnustep-1.7` by calling:

```cmake
addtest_flags("${TEST}_legacy" "-O0 -fobjc-runtime=gnustep-1.7 -UNDEBUG" "${TEST_SOURCE}")
```

which will result in the same flag being defined twice (`-fobjc-runtime=gnustep-2.0 -fobjc-runtime=gnustep-1.7`).  In my setup, that resulted in clang targeting the 2.0 runtime, making the legacy tests invalid.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,7 +175,6 @@ set(LIBOBJC_NAME "objc" CACHE STRING
 set(INCLUDE_DIRECTORY "objc" CACHE STRING 
 	"Subdirectory of the include path to install the headers.")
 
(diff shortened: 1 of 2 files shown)


davidchisnall:
Interesting. This should be a last-version-wins flag in clang.

davidchisnall (approved):

davidchisnall:
Looks like FreeBSD CI is failing because the versions that we’re testing on are EOL.

davidchisnall:
[ Updating the branch to pick up fixes to the CI scripts. ]

PR #256, opened 2023-12-31
Update FreeBSD versions

qmfrederik:
Match latest versions at https://www.freebsd.org/releases/, 12.4 went EOL on 31-dec-23: https://www.freebsd.org/security/#sup


davidchisnall:
Sorry, didn't see that you were doing this as well.

PR #262, opened 2024-01-01
Add MinGW CI

qmfrederik:
Adds a CI job which builds libobjc2 in the the msys2/ucrt64 and the msys2/mingw64 environments.  I had to disable a couple of tests for CI to pass; we can re-enable most of them after/when #260 is merged.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -188,6 +188,48 @@ jobs:
       run: |
         ctest -j 4 --output-on-failure -T test
 
+  mingw:
+    strategy:
+      matrix:
+        # Build each combination of OS and release/debug variants
+        os: [ windows-2019 ]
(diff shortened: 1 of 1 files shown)


davidchisnall:
Hopefully more of the tests are passing now?

qmfrederik:
Yes, all tests but `UnexpectedException` pass on MinGW, but we'd need a way for CI to pick up the latest version of clang and/or allow the exception tests to fail until we've got clang 18 in MinGW.

qmfrederik:
The clang changes to support compiling Objective-C code on MinGW have been backported onto the version of clang which ships with MinGW (see https://github.com/msys2/MINGW-packages/pull/19767).  This means CI now works on MinGW, including the exception handling tests.

@davidchisnall Let me know if this is OK with you.

davidchisnall (approved):

PR #263, opened 2024-01-02
Skip tests using objc_msgSend if not available

hmelder:
This PR fixes compilation of unit tests when objc_msgSend is not available.


Test/ManyManySelectors.m
```diff
@@ -49,5 +50,6 @@ int main(void)
 	objc_msgSend([Test class], nextSel);
 	assert(methodCalled == YES);
 	return 0;
+#endif // __GNUSTEP_MSGSEND__
```
davidchisnall, line 53: This test uses both lookup paths, it would be good to test the one that works everywhere unconditionally

Test/CMakeLists.txt
```diff
@@ -124,6 +124,7 @@ function(addtest_flags TEST_NAME FLAGS TEST_SOURCE)
 		LINK_FLAGS ${INCREMENTAL}
 		LINKER_LANGUAGE C
 	)
+	set_property(TEST ${TEST_NAME} PROPERTY SKIP_RETURN_CODE 77)
```
davidchisnall, line 127: What does this do?

hmelder: If the return value of the test is `77`, the test is marked as skipped.
https://www.scivision.dev/cmake-skip-test-code-77/

Test/ManyManySelectors.m
hmelder: I see

Test/CMakeLists.txt
davidchisnall: Oh, interesting.  For other tests that don't work in some configurations, we just don't run them and rely on the number of executed tests being smaller than the total set.

hmelder: The thing is that it is easier run a unit test and return `77` based on a condition, then do a test in CMake if `__GNUSTEP_MSGSEND__` is set. We *could* streamline platform detection by moving everything into CMake configuration.

hmelder: > we just don't run them 

Has the disadvantage of not seeing if a test was skipped.

davidchisnall: Yup, I agree that this is better, just wondering if we should do the same thing with other tests...

hmelder:
The `UnexpectedException` test is now compiled, but skipped on arm and AArch64.

I have rebased the branch and added a small test which detects errors like #265.

davidchisnall (approved):

PR #266, opened 2024-01-04
AArch64 objc_msgSend: Fix argument corruption due to invalid stack pointer offset

### LINKED ISSUE #265: Wrong stack pointer offset in objc_msgSend.aarch64.S leads to register corruption
I was working on a test to fix this and stumbled upon #264. This was not trivial to find, as it was causing some random exceptions in libs-base.

```patch
diff --git a/objc_msgSend.aarch64.S b/objc_msgSend.aarch64.S
index 25cbfd6..e9e30ba 100644
--- a/objc_msgSend.aarch64.S
+++ b/objc_msgSend.aarch64.S
@@ -169,7 +169,7 @@ CDECL(objc_msgSend_stret):
        EH_NOP
        ldp    x2, x3, [sp, #32]
        EH_NOP
-       ldp    x4, x5, [sp, #64]
+       ldp    x4, x5, [sp, #48]
        EH_NOP
        ldp    x6, x7, [sp, #64]
        EH_NOP
```

I'll make a PR with this fix and a unit test.

hmelder:
Fixes #265 

We can run the tests with `-Xclang -fobjc-dispatch-method=non-legacy`, but we would need to check for fast-path support in the CMake configuration first.

### DIFF
--- a/Test/CMakeLists.txt
+++ b/Test/CMakeLists.txt
@@ -134,9 +134,9 @@ function(addtest_flags TEST_NAME FLAGS TEST_SOURCE)
 endfunction(addtest_flags)
 
 function(addtest_variants TEST TEST_SOURCE LEGACY)
-	addtest_flags(${TEST} "-O0 -fobjc-runtime=gnustep-2.0 -UNDEBUG -DGS_RUNTIME_V2" "${TEST_SOURCE}")
+	addtest_flags(${TEST} "-O0 -fobjc-runtime=gnustep-2.2 -UNDEBUG -DGS_RUNTIME_V2" "${TEST_SOURCE}")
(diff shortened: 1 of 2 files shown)


davidchisnall (approved):

davidchisnall (approved):

triplef:
What does using `-fobjc-runtime=gnustep-2.2` instead of `2.0` do in Clang? Is that something that we should update in other projects using the runtime as well?

davidchisnall:
> What does using `-fobjc-runtime=gnustep-2.2` instead of `2.0` do in Clang? Is that something that we should update in other projects using the runtime as well?

In the released versions of clang, nothing.  In the next release, it will enable the fast-path dispatch for AArch64.

PR #273, opened 2024-01-21
Fix offsets in selector_table

hmelder:
Seems like the NUL check and offsets are wrong here.

### DIFF
--- a/selector_table.cc
+++ b/selector_table.cc
@@ -172,7 +172,7 @@ static BOOL selector_types_equal(const char *t1, const char *t2)
 {
 	if (t1 == nullptr || t2 == nullptr) { return t1 == t2; }
 
-	while (('\0' != *t1) && ('\0' != *t1))
+	while (('\0' != *t1) && ('\0' != *t2))
 	{
 		t1 = skip_irrelevant_type_info(t1);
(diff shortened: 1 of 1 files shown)


davidchisnall (approved):

triplef:
Good finds! Can you comment on what kinds of issues this might have caused at runtime?

hmelder:
> Can you comment on what kinds of issues this might have caused at runtime?

This one should only be relevant, if you are using GCC (or an object file compiled using GCC), as it generates "*" for @encode(BOOL*), while Clang generates "^c" or "^C". When comparing type encodings, we check for both options.
```diff
		if ((*t1 == '*') && (*t2 != '*'))
		{
- 			if (*t2 == '^' && (((*(t2+1) == 'C') || (*(t2+2) == 'c'))))
+			if (*t2 == '^' && (((*(t2+1) == 'C') || (*(t2+1) == 'c'))))
```

Here we do not stop if the second string ends, but as the pointer increment is guarded by an additional check (See line 216), we keep comparing the next char from t1 with `\0` until t1 is also `\0`.
```diff
-	while (('\0' != *t1) && ('\0' != *t1))
+	while (('\0' != *t1) && ('\0' != *t2))
```

PR #274, opened 2024-02-05
CMakeLists.txt: fixed mistyped name `tsl-robin-map_FOUND` variable.

trunkmaster:
This change let package maintainers use `tsl-robin-map_DIR` cmake parameter with path set to cloned and built robin-map repo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,7 @@ endif ()
 
 find_package(tsl-robin-map)
 
-if (NOT tls-robin-map_FOUND)
+if (NOT tsl-robin-map_FOUND)
 	FetchContent_Declare(
 		robinmap
 		GIT_REPOSITORY https://github.com/Tessil/robin-map/


davidchisnall (approved):

davidchisnall:
Great catch, thanks!  I wonder if we should do a 2.2.1 to incorporate this.  I completely missed it in review.

davidchisnall:
FYI: Part of the motivation for this change (modulo the bug that you've fixed) is that package maintainers should be able to use their system-installed version, they shouldn't need to set `tsl-robin-map_DIR` as long as CMake's search path for packages includes the system install location.

trunkmaster:
> FYI: Part of the motivation for this change (modulo the bug that you've fixed) is that package maintainers should be able to use their system-installed version, they shouldn't need to set `tsl-robin-map_DIR` as long as CMake's search path for packages includes the system install location.

Not exactly. I mean cloning robin-map github repo, configure, build and after that specify -Dtsl-robin-map_DIR=<directory/where/robin-map/built>. I don't need to install robin-map in that case.

qmfrederik:
> Not exactly. I mean cloning robin-map github repo, configure, build and after that specify -Dtsl-robin-map_DIR=<directory/where/robin-map/built>. I don't need to install robin-map in that case.

The build system should do that for you through `FetchContent`.  So you shouldn't have to manually clone that repo, unless you're looking for a specific version of robin-map?

qmfrederik:
That makes sense, thanks for the clarification!

PR #277, opened 2024-02-12
MinGW: Remove manual setting of library prefix/suffix

qmfrederik:
We no longer need to manually set these suffixes/prefixes as CMake in MSYS2 now knows about Objective C.

See https://github.com/msys2/MINGW-packages/pull/20028
See https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9244
See https://github.com/msys2/MINGW-packages/pull/20024#issuecomment-1938493481

/cc @MehdiChinoune


MehdiChinoune:
I don't where does the issue come from, those are set by default by CMake, they shouldn't be added at all.

qmfrederik:
My best guess is that this is because the CMake project [declares itself as a OBJC/OBJCXX project](https://github.com/gnustep/libobjc2/blob/master/CMakeLists.txt#L21), and CMake may not have sensible defaults for Objective-C on MSYS2/MINGW and friends.

MehdiChinoune:
> My best guess is that this is because the CMake project [declares itself as a OBJC/OBJCXX project](https://github.com/gnustep/libobjc2/blob/master/CMakeLists.txt#L21), and CMake may not have sensible defaults for Objective-C on MSYS2/MINGW and friends.

We should add support for MinGW for OBJC/OBJCXX to CMake.
I will try to patch CMake and upstream.

MehdiChinoune:
@qmfrederik Do you know which clang library has objc_alloc, objc_alloc_init and objc_allocWithZone symbols?

davidchisnall:
> Do you know which clang library has objc_alloc, objc_alloc_init and objc_allocWithZone symbols?

These are not provided by clang, they are provided by libobjc2.

MehdiChinoune:
> > Do you know which clang library has objc_alloc, objc_alloc_init and objc_allocWithZone symbols?
> 
> These are not provided by clang, they are provided by libobjc2.

I am trying to support OBJ/OBJCXX on CMake, but It fails with undefined reference to those symbols!
Looks like they weren't exported.

MehdiChinoune:
> @MehdiChinoune Have you got a small repo? libobjc2 does export those symbols. You also need to make sure you use the right clang parameters -- e.g. `-fobjc-runtime=gnustep-2.0` if you you use the libobjc2.0 runtime.

It's about this repo, I am testing your on MinGW-package

davidchisnall:
Those symbols won’t be exported from anywhere until libobjc2 is built…

qmfrederik:
@MehdiChinoune Thanks for your patches for cmake (https://github.com/msys2/MINGW-packages/pull/20028).  I've removed the the hard-coded values for `CMAKE_IMPORT_LIBRARY_SUFFIX`, `CMAKE_LINK_LIBRARY_SUFFIX`, and `CMAKE_SHARED_LIBRARY_PREFIX` and the the shared library and linker library now get installed as `bin/libobjc.dll` and `lib/libobjc.dll.a`.

Can you confirm this is what you had in mind?

MehdiChinoune:
Yes, fixing shared/import libraries suffix/prefix was the intention.
before you merge this PR, could you try setting `LINKER_LANGUAGE` (https://cmake.org/cmake/help/latest/prop_tgt/LINKER_LANGUAGE.html) to `CXX` and see if it fixes the required LDFLAGS.

qmfrederik:
@MehdiChinoune I tried:
- Setting `_DLINKER_LANGUAGE=CXX` but that variable is not used: "CMake Warning: Manually-specified variables were not used by the project: LINKER_LANGUAGE"
- I thought I could perhaps pull in stdc++ and gcc_s into ObjC projects by setting `CMAKE_OBJCXX_IMPLICIT_INCLUDE_DIRECTORIES` to `CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES`. That causes the code to link with a bunch of other libraries, too (including `mingwex` and `mingw32`) which causes the exception handling tests to fail

Long story short, I would leave it at this for now -- we can always revisit later and try to optimize things.

davidchisnall (approved):
With the caveat that I don't have a Windows machine to try this on, it looks fine.

davidchisnall:
Build looks green for MinGW CI, and I'm in favour of removing complexity from the build system if it is not essential.

MehdiChinoune:
> @MehdiChinoune I tried:
> - Setting `_DLINKER_LANGUAGE=CXX` but that variable is not used: "CMake Warning: Manually-specified variables were not used by the project: LINKER_LANGUAGE"

LINKER_LANGUAGE is a target property, It couldn't be passed by command.

qmfrederik:
@MehdiChinoune Do you know which clang library has objc_alloc, objc_alloc_init and objc_allocWithZone symbols?

You were right, they weren't being exported on Windows. I've included a fix for that in #280.

PR #280, opened 2024-02-21
MinGW: Add exception-handling façades to libobjc2

qmfrederik:
The current implementation uses functions which are declared in the 
C++ standard library to throw and catch exceptions.
This requires linking with said library, which most toolchains do not
do natively.

### DIFF
--- a/fast_paths.m
+++ b/fast_paths.m
@@ -12,6 +12,7 @@ typedef struct _NSZone NSZone;
 /**
  * Equivalent to [cls alloc].  If there's a fast path opt-in, then this skips the message send.
  */
+OBJC_PUBLIC
 id
 objc_alloc(Class cls)
 {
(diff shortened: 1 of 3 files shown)


qmfrederik:
Also fixes the visibility of `objc_alloc`, `objc_allocWithZone` and `objc_allocInit` - https://github.com/gnustep/libobjc2/pull/277#issuecomment-1939214893

qmfrederik:
Also removes the weak annotation on functions in `objcxx_eh.h` - https://github.com/gnustep/libobjc2/pull/279/files#r1492653201

qmfrederik:
clang fix: https://github.com/llvm/llvm-project/commit/10db940979c2753ef547956e6b94f610df6a84fe

qmfrederik:
@davidchisnall Would you mind reviewing this and let me know what you think?

davidchisnall (approved):

Issue #284, opened 2024-03-08
Rare class init / atomic property deadlock


ERobsham:
While investigating some intermittent deadlocks being observed in a processes that is part of a suite of software I help maintain, I think I ended up tracking down a potential deadlock. 

(Partially due to how pointer hashes could collide for the for the 'spinlocks' used in a few places... but I'm not sure _any_ hash algorithm would ensure this _couldn't_ be hit, just help reduce the rate it happens at)

The process seems to _only_ lock up during initial startup. If gets through initial startup, it seems like it continues chugging along without any hiccups.  The following rabbit hole I'm about to describe does use this fact as a basis for making some assumptions on which code paths are taken.

----

Stack trace of the deadlocked thread (this is a multi threaded application, but none of the other threads are really interesting to look at):
```
Thread 1 (Thread 24814.24814):
#0  0x0000ffffa4ba0120 in clock_nanosleep () 
#1  0x0000ffffa4ba570c in nanosleep () 
#2  0x0000ffffa4ba55e4 in sleep () 
#3  0x0000ffffa4d4d9e0 in ?? ()
#4  0x0000ffffa4d4dec4 in objc_sync_enter ()
#5  0x0000ffffa4d3e6dc in ?? ()
#6  0x0000ffffa4d46b34 in objc_msg_lookup_sender ()
#7  0x0000ffffa4d4bfe0 in objc_retain ()
#8  0x0000ffffa4d4eb64 in objc_getProperty ()
#9  0x0000ffffa14a2fa0 in -[SomeObj start] (self=0xaaaafd5ab608, _cmd=<optimized out>)
... remaining trace isn't very relevant
```

So it looks like inside `objc_getProperty` we attempt to `retain`, while inside a `spin_lock(0xaaaafd5ab608 + propOffset)` 
```c
OBJC_PUBLIC id objc_getProperty(id obj, SEL _cmd, ptrdiff_t offset, BOOL isAtomic) {
	// ...
    char *addr = (char*)obj;
	addr += offset;
    // ...
	if (isAtomic) {                                 // in our case, `isAtomic` is `true`
		volatile int *lock = lock_for_pointer(addr);
		lock_spinlock(lock);                        // locked on lock_for_pointer(obj + offset) 
		ret = *(id*)addr;
		ret = objc_retain(ret);                     // <==== we're inside here @ bt #7
	    // ...
```

**_At this point we've taken the `lock_spinlock(0xaaaafd5ab608 + propOffset)`_**  

[...]

PR #285, opened 2024-03-08
Potential atomic property/init deadlock fix

ERobsham:
Initial attempt at resolving issue #284 (Rare atomic property during initialization deadlock)

### DIFF
--- a/associate.m
+++ b/associate.m
@@ -12,6 +12,8 @@
 #include "lock.h"
 #include "gc_ops.h"
 
+PRIVATE int associate_spinlocks[spinlock_count];
+
 /**
  * A single associative reference.  Contains the key, value, and association
  * policy.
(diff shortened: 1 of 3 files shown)


davidchisnall:
I think this code looks correct, it would be nice to have a test.  I’m still not quite sure how code can get to this path though.

ERobsham:
That _is_ a tough one.   I'm not really sure how to _force_ this behavior, as its been very intermittent in our systems.   There are definitely a couple of exacerbating factors I know of:

- initializing objects with _a lot_ of atomic properties, then calling an instance methods on those atomic properties.
- running on `aarch64` (seems like the pointer hashes are much more likely to collide).

The process that is deadlocking for us has the 'main application' class with a ton of atomic properties, and most of those classes are also host to a bunch of atomic properties, in some cases this continues a number of levels deep...

As far as I can tell the only way to hit this is more or less a numbers game.   You need to have an instance of an object where it has an atomic property, and both `theAtomicProp` and `theAtomicProp.class->isa` produce the same hash when run through the `lock_for_pointer()` hashing -- **_And_** the class for the atomic property needs to currently have an uninstalled dtable when you try to 'get' the property for the first time.

davidchisnall:
I believe this is now fixed.  Sorry for ignoring it for so long!

PR #287, opened 2024-03-27
Generate CMake Configuration File

qmfrederik:
Installs:
- lib/cmake/libobjc/libobjcConfig.cmake
- lib/cmake/libobjc/libobjcTargets.cmake
- lib/cmake/libobjc/libobjcConfigVersion.cmake
- lib/cmake/libobjc/libobjcTargets-noconfig.cmake

So that downstream projects can use `find_package(libobjc)` to link with libobjc2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 *~
 .*.sw?
-Build
+[b|B]uild
 Debug
 Release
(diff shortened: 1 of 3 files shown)


davidchisnall (approved):
Thanks!

PR #288, opened 2024-04-02
Don't use CXXFLAGS when compiling eh_trampoline.cc

### LINKED ISSUE #177: Can't link certain programs with libobjc: undefined reference: eh_trampoline()

qmfrederik:
We don't want the user to override these flags because that file needs to be compiled in a very specific way.

Re-applies #178 (365e53632e8be41e49f21ee47a63e41be424a237)
Related to #183
Fixes #177


qmfrederik:
The context here is that compilation currently fails when trying to build a .rpm package for RedHat; I assume that process sets additional CXX_FLAGS which mess with the custom command which generates `eh_trampoline.S`.

Looking at #183, I'm not sure if we need to re-introduce CMAKE_CXX_FLAGS in EH_PERSONALITY_FLAGS when CMAKE_CXX_COMPILER_TARGET is set?

davidchisnall (approved):

PR #290, opened 2024-04-15
Support Control Flow Enforcement with Indirect Branch Tracking

buzzdeee:
This is enforced on OpenBSD platforms/hardware that supports it, namely Intel gen11 or newer (amd64), or Apple M2 (aarch64).

Patch is needed to let GNUstep applications work on such machines. I don't have such "modern" hardware, but some minimal tests at least on an amd64 box that supports it done. aarch64 may need more to make it work properly.

More see threat: https://marc.info/?t=170974826400001&r=1&w=2

### DIFF
--- a/block_trampolines.S
+++ b/block_trampolines.S
@@ -23,6 +23,7 @@
 // x86-64 trampoline
 ////////////////////////////////////////////////////////////////////////////////
 .macro trampoline arg0, arg1
+	endbr64
 	mov   -0x1007(%rip), \arg1   # Load the block pointer into the second argument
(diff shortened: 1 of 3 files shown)


davidchisnall:
The changes to be blocks trampolines look like they'll make the trampolines larger than two pointers and so require some changes to the size typedefs.  I'm somewhat unconvinced by CFI for Objective-C because it's trivial to overwrite an on-stack block invoke pointer and build a Turing-complete weird machine out of valid branch targets, but I don't object to merging this if it doesn't break anything.  For some reason, this didn't run the GitHub Actions, which would have tested on AArch64.  The message-send changes look fine.

PR #291, opened 2024-04-26
Whitelist UnexpectedException test for all architectures and recommend libunwind in INSTALL

hmelder:
I am undecided about whether to add a try_compile to the CMakeLists.txt to automatically use CompilerRT and libunwind if they are available. This should exclude Windows and Android.

This PR should be merged after a fix for [libgcc: _Unwind_RaiseException corrupts return value](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114843) is implemented in GCC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 [b|B]uild
 Debug
 Release
+.cache
--- a/INSTALL
+++ b/INSTALL
@@ -6,6 +6,14 @@ runtime.  Note that although this runtime should build with gcc, it is strongly
 recommended that you compile it with clang and clang++ as your [Objective-]C and
 [Objective-]C++ compilers.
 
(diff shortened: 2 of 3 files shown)


hmelder:
On the other hand, automatic configuration is better than manually setting compile flags.

hmelder:
What do you think @davidchisnall?

davidchisnall:
I think this is the right thing (implicitly linking a non-default unwinded will break things for people), but I’d like CI to be green…

hmelder:
>  but I’d like CI to be green…

We can use CompilerRT + libunwind in the CI

hmelder:
I guess finding prebuild libunwind cross builds for ubuntu might be a bit difficult

hmelder:
Seems like the GCC patch has not arrived in 24.04LTS :/

PR #296, opened 2024-07-23
Android CI: Enable KVM, Caching and AArch64

hmelder:


.github/scripts/android_test_driver.sh
```diff
@@ -15,26 +15,28 @@ for BINARY in $BINARIES; do
     TOTAL=$((TOTAL + 1))
     
     START_TIME=$(date +%s)
-    START_TIME_MS=$((START_TIME * 1000 + $(date +%N) / 1000000))
+    # Busybox date does not support %N, so we can't get milliseconds this way
```
davidchisnall, line 18: Are we using Busybox or bash here?  The change at the top implies the latter.

hmelder: > Are we using Busybox or bash here?

The shebang change is wrong. Correct.

davidchisnall (approved):

PR #298, opened 2024-08-26
Remove Legacy ABI Compatibility

hmelder:
We are carrying a lot of dead-weight just by supporting the old GCC ABI which probably no one uses in combination with libobjc2.

There are still lot more ugly hacks which can be removed.


legacy.c
```diff
@@ -322,8 +322,8 @@ static struct objc_property_list *upgradePropertyList(struct objc_property_list_
 	{
 		return NULL;
 	}
-	size_t data_size = l->count * sizeof(struct objc_property);
-	struct objc_property_list *n = calloc(1, sizeof(struct objc_property_list) + data_size);
+	size_t ARRAY_DATA_SIZE = l->count * sizeof(struct objc_property);
```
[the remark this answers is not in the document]
hmelder: Ups this was a mistake. I replaced `data_size` in sarray and accidentally replaced this one as well.

hmelder:
> Isn’t there more to be deleted from the upgrade logic?

Yeah but I was not sure how much to delete. Do we want to leave the gsv1 upgrade logic in place or remove it as well?

hmelder:
And this would fall under `OLDABI_COMPAT` right?

davidchisnall:
> > Isn’t there more to be deleted from the upgrade logic?
> 
> Yeah but I was not sure how much to delete. Do we want to leave the gsv1 upgrade logic in place or remove it as well?

I'm not sure.  We currently don't support the new ABI on Mach-O in clang and it's quite useful to be able to run tests on macOS so that we can compare their behaviour against the Apple runtime.  I've not tried WebAssembly, but it probably also needs some custom work.

The main reason I'd drop the GCC ABI is that anyone using GCC should probably use their runtime.  Ideally we'd get every target supported by the v1 GNUstep ABI onto v2 and then drop v1 as well.

hmelder:
> We currently don't support the new ABI on Mach-O in clang 

I did not know that we support v1 on Macho-O. It would be very interesting to support v2 as well just for benchmarking purposes.

hmelder:
I'll remove the GCC upgrade logic then.

hmelder:
Google recently updated the clang prebuilds for the Android SDK. The fast path test now executes and fails instead of being skipped.

hmelder:
@davidchisnall can you take another look at it. I've rebased it onto the latest changes in master.

selector_table.cc
```diff
@@ -294,20 +263,11 @@ struct SelectorHash
 			hash = hash * 33 + c;
 		}
 #ifdef TYPE_DEPENDENT_DISPATCH
-		// We can't use all of the values in the type encoding for the hash,
-		// because our equality test is a bit more complex than simple string
-		// encoding (for example, * and ^C have to be considered equivalent, since
-		// they are both used as encodings for C strings in different situations)
```
davidchisnall, line 300: Are you sure clang never does this?  I vaguely remember things like `char*` and `[un]signed char*` encoded differently.

selector_table.cc
```diff
@@ -294,20 +263,11 @@ struct SelectorHash
...
 		if ((str = types))
 		{
 			while((c = (size_t)*str++))
 			{
-				switch (c)
-				{
-					case '@': case 'i': case 'I': case 'l': case 'L':
-					case 'q': case 'Q': case 's': case 'S': 
-					hash = hash * 33 + c;
-				}
```
davidchisnall, line 310: I'm not sure if this change is correct with respect to structs and so on.

PR #303, opened 2024-10-21
Fast Paths: Check if argument is nil

### LINKED ISSUE #302: Crashes in fast_paths.m when class argument is null
The subject says it ... the functions in fast_paths.m don't check for a null argument and hence crash when dereferencing the null pointer.
It's easy to demonstrate with a pattern like:

Class
getClassNamed(char *name)
{
  return Nil;
}

id obj = [[getClassNamed("dummy") alloc] init];

hmelder:
Fixes #302

### DIFF
--- a/fast_paths.m
+++ b/fast_paths.m
@@ -16,6 +16,10 @@ OBJC_PUBLIC
 id
 objc_alloc(Class cls)
 {
+	if (UNLIKELY(cls == nil))
+	{
+		return nil;
+	}
 	if (UNLIKELY(!objc_test_class_flag(cls->isa, objc_class_flag_initialized)))
 	{
 		objc_send_initialize(cls);
@@ -34,6 +38,10 @@ OBJC_PUBLIC
 id
 objc_allocWithZone(Class cls)
 {
+	if (UNLIKELY(cls == nil))
+	{
+		return nil;
+	}
 	if (UNLIKELY(!objc_test_class_flag(cls->isa, objc_class_flag_initialized)))
 	{
 		objc_send_initialize(cls);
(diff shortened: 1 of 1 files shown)


davidchisnall:
Fix looks good, can we have a test? Call these methods with `objc_getClass(“flibble”)` and make sure they work?

hmelder:
Turns out the bug is not triggered when using `objc_getClass("flibble")` as the receiver:
```
56:                                               ; preds = %54
  %57 = call ptr @objc_getClass(ptr noundef @.str.12)
  %58 = call ptr @objc_msgSend(ptr noundef %57, ptr noundef @".objc_selector_alloc_\0116\010:8"), !GNUObjCMessageSend !8
  %59 = call ptr @objc_getClass(ptr noundef @.str.12)
  %60 = call ptr @objc_msgSend(ptr noundef %59, ptr noundef @".objc_selector_alloc_\0116\010:8"), !GNUObjCMessageSend !8
  %61 = call ptr @objc_msgSend(ptr noundef %60, ptr noundef @".objc_selector_init_\0116\010:8"), !GNUObjCMessageSend !9
```

 However, this works:

```objc
Class
getClassNamed(char *name)
{
return nil;
}
```



```
56:                                               ; preds = %54
  %57 = call ptr @getClassNamed(ptr noundef @.str.12)
  %58 = call ptr @objc_alloc(ptr %57)
  %59 = call ptr @getClassNamed(ptr noundef @.str.12)
  %60 = call ptr @objc_alloc_init(ptr %59)
 ```

davidchisnall (approved):

PR #309, opened 2024-11-14
Windows: Support building with strict compatibility

qmfrederik:
This commit adds the `STRICT_APPLE_COMPATIBILITY` CMake option which will force `BOOL` to defined as an `signed char` on Windows instead of an `int`, as having different types for `BOOL` on different platforms creates serialization issues.

On MSYS, the `BOOL` type is defined in` minwindef` like this:

```
#if !defined(__OBJC__) && !defined(__OBJC_BOOL) && !defined(__objc_INCLUDE_GNU) && !defined(_NO_BOOL_TYPEDEF)
  typedef int BOOL;
#endif
```

To prevent this header from defining a conflicting `BOOL` type, this commit also defines `__OBJC_BOOL`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -213,6 +213,7 @@ jobs:
         os: [ windows-2019 ]
         msystem: [ ucrt64, mingw64, clang64 ]
         build-type: [ Release, Debug ]
+        strict-apple-compatibility: [ ON, OFF ]
         include:
           - msystem: ucrt64
             package-prefix: ucrt-x86_64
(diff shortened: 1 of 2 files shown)


qmfrederik:
@davidchisnall You introduced defining `BOOL`  as `int` on Win32 in 934cd13a7a346dc0f3d8466fd180a127354a0ee5.  Can you confirm that's the right approach? GCC always uses `unsigned char`.

qmfrederik:
> I think we made BOOL match the type of the same name in the Windows headers. Having different definitions caused problems. 

Yes, I agree that's certainly a world of pain.  MSYS seems to have done a good job at consistently using `WINBOOL` instead of `BOOL`.  I don't think this would work if you compile on Windows using a native toolchain (clang or msvc) as the headers in the Windows SDK don't offer that flexibility.

Having said all that -- this PR makes it an opt-in feature, and you'd have to configure the CMake project with `-DSTRICT_APPLE_COMPATIBILITY`.  So we let the user decide.

So if you're good with this opt-in behavior, let me know and we can get this merged.

davidchisnall (approved):

PR #313, opened 2024-11-17
Fix selector forwarding on x64 Windows

### LINKED ISSUE #312: Selector forwarding is broken on Windows (mingw and msvc)
Selector forwarding from one object to another seems to be broken when using libobjc2 on Windows (see also: https://github.com/gnustep/plugins-themes-WinUXTheme/pull/7#issuecomment-2480839795).

I've opened a PR in libs-base which shows the problem: https://github.com/gnustep/libs-base/pull/466

qmfrederik:
As described in #312, selector forwarding does not appear to be working correctly on Windows.  This becomes apparent when accessing `[self]`.  This PR updates the Forward test to access an instance variable, exposing the issue, and proposes a fix.

The root cause appears to be that the `rcx` register is being overwritten after `slowMsgLookup` returns.  On Unix, `rcx` is the fourth argument; on Windows, it is the first argument. Introduce a `FOURTH_ARGUMENT` variable to ensure we use `rcx` on Unix and `r9` on Windows to refer to the fourth argument.

Closes #312

### DIFF
--- a/Test/Forward.m
+++ b/Test/Forward.m
@@ -18,14 +18,24 @@ id target;
 @end
 
 @interface ForwardingTarget : Test
+{
+  int y;
+}
 @end
 
 BOOL forwardingTargetCalled;
 
 @implementation ForwardingTarget
+- (id)init
+{
+	y = 42;
+	return self;
+}
+
 - (void)foo: (int)x
 {
 	assert(x == 42);
+	assert(x == y);
 	forwardingTargetCalled = YES;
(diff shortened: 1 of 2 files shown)


qmfrederik:
@davidchisnall This is the simples repro I could create so far.  Looks like this issue happens on 64-bit Windows only; 32-bit Windows appears to be fine.

davidchisnall (approved):

PR #314, opened 2024-11-18
Persist STRICT_APPLE_COMPATIBILITY in objc-config.h

qmfrederik:
On Windows, the value of `STRICT_APPLE_COMPATIBILITY` determines whether `BOOL` is a char or an int.  This value must be consistent across all libraries that use libobjc2.  Rather than requiring all libraries to explicitly define `STRICT_APPLE_COMPATIBILITY`, persist the value of `STRICT_APPLE_COMPATIBILITY` in `objc-config.h` and include that in `runtime.h`.


davidchisnall (approved):

davidchisnall:
Might be good to propagate the other build-time configuration things there.

qmfrederik:
@davidchisnall Sure, which ones would you like?  Some of them have generic names (e.g. `LEGACY_COMPAT`) so perhaps we should prefix them with OBJC_?

qmfrederik:
I'll merge as is but we can always add more configuration changes later.

PR #318, opened 2024-11-18
msgSend: Return `nil` if the dtable cannot be loaded

qmfrederik:
Under some circumstances, apparently when two classes with the same name exist in two different modules, the dtable for a class can be null.  This would cause a crash in `msgSend`.

Update `msgSend` to return `nil` instead.

### DIFF
--- a/objc_msgSend.x86-64.S
+++ b/objc_msgSend.x86-64.S
@@ -35,6 +35,8 @@
 	mov   (\receiver), %r10               # Load the dtable from the class
 1:	                                      # classLoaded
 	mov   DTABLE_OFFSET(%r10), %r10       # Load the dtable from the class into r10
+	test  %r10, %r10                      # If the dtable is nil
+	jz    4f                              # return nil
 	mov   %rax, -8(%rsp)                  # %rax contains information for variadic calls
 	mov   %rbx, -16(%rsp)                 # On the fast path, spill into the red zone
 	mov   (\sel), %eax                    # Load the selector index into %eax


qmfrederik:
Here's an example of GNUstep including the `GSColorSliderCell` class in two separate modules/bundles: https://github.com/gnustep/libs-gui/blob/gui-0_31_1/ColorPickers/GNUmakefile#L47-L49.  Loading a `NSColorPanel` would cause a crash without this change when using libobjc2+clang.  Seems to work fine when using GCC.

qmfrederik:
You should be able to reproduce this with the [GSTest](https://github.com/gnustep/tests-examples/tree/master/gui/GSTest) app in the [gnustep/tests-examples](https://github.com/gnustep/tests-examples) repository; without this change, you'll get a segfault when clicking **Format**, **Font**, **Colors...**.

hmelder:
Why not use class references for this?

Here, the shared library `unit2.m` contains the a ref to `Foo` which was defined in the main program `unit1.m`. `foo.h` contains the interfaces for Foo and SomeOtherClass.

## unit2.m
```objc
#import "foo.h"
#include<stdio.h>

@class Foo;

@implementation SomeOtherClass

+ (void) load {
	puts("Loaded from unit2!");
}

- (void) doStuffStuff {
	[[Foo new] doStuff];
}

- (void) doStuffWithFoo: (Foo *) f {
	[f doStuff];
}

@end
```

Compiled using 
```clang unit2.m `gnustep-config --objc-flags` `gnustep-config --base-libs` -shared -o libunit2.so```

## unit1.m
``` objc
#import "foo.h"
#include <stdio.h>

@implementation Foo

- (void) doStuff {
	puts("stuff!");
}

@end

int main(void) {
 	SomeOtherClass *obj = [SomeOtherClass new];
	[obj doStuffStuff];
	[obj doStuffWithFoo: [Foo new]];
}
```

Note that I am not loading the shared library at runtime, but this should not make a difference if the NSBundle loading code in libs-base is correct.

````clang unit1.m -lunit2 -L./ `gnustep-config --objc-flags` `gnustep-config --base-libs````


```sh
hugo@ubuntu:~/class_refs$ LD_LIBRARY_PATH=./ ./a.out
Loaded from unit2!
stuff!
stuff!
```

No duplication of the actual implementation.

qmfrederik:
I agree on this being undefined behavior.  The question is, what should the runtime do?  In particular in this case, where the same class exists in two modules, but it is the exact same class (same source code file being compiled into two modules).

Currently, it causes an access violation in a macro in assembly code.  That's probably the worst place to error out, as it makes it very difficult to troubleshoot.

This code has been in gnustep-gui for 17 years (https://github.com/gnustep/libs-gui/commit/f22c862811aaf11bb4d23da836c3765f2edb83f6) and has been working for the past 17 years.

So perhaps it's worth for libobjc2 to (try to) handle this gracefully?  I may be addressing a symptom here and there may be a better place to handle this, though?

davidchisnall:
I don’t see a way of fixing this without an additional branch in the message send function. This is the hot path for most Objective-C programs and so every extra instruction there can have a measurable impact on performance. A branch that’s always not-taken might be fine, but it’s not great. 

I believe the Apple runtime gives a linker failure in these cases.

hmelder:
> Are there any (micro)benchmarks I could use to quantify the impact? If not, a framework you'd recommend?

Google Test is a good start. For scheduling analysis, [llvm-mca](https://llvm.org/docs/CommandGuide/llvm-mca.html) is great but heavily depends on the quality of the CPU's Scheduling Model and does not simulate caching.

davidchisnall:
We don’t have good performance testing, but using sample-based profilers you can see how much time is spent in `objc_msgSend` in a codebase you care about.

triplef:
Side note: can we in general integrate that benchmark into CI to ensure future changes don’t impact performance in a negative way? For our app a 7% performance impact could have been fairly significant.

qmfrederik:
Benchmark signals on CI machines may be a bit noisy, but what I've done in other projects is to add two steps to a CI pipeline:
- Build and run the benchmark on the current branch
- Build and run the benchmark on the main branch

That at least gives some kind of signal w.r.t. any changes in the performance characteristics being introduced.  I can open a separate PR for that, if there's consensus that the benchmark is useful (and everyone is happy with using Google's benchmark framework)

rfm:
For what it's worth, I think benchmarking is very valuable as long as the margins aren't set so tight that we get lots of false positives due to other activity on the machine.

qmfrederik:
Closing this for now. See #319 for the benchmarks

PR #323, opened 2024-12-20
Fix leak of weak references

rfm:
In the case where an object has been deallocated, fetching the old value of a weak reference to it returns nil.  If we are assigning nil to the weak reference we can't abort processing early (when we check that old and new values are the same) as that will leave the weak reference in place when the calling code thinks it has destroyed it.

### DIFF
--- a/arc.mm
+++ b/arc.mm
@@ -838,8 +838,9 @@ extern "C" OBJC_PUBLIC id objc_storeWeak(id *addr, id obj)
 	WeakRef *oldRef;
 	id old;
 	loadWeakPointer(addr, &old, &oldRef);
-	// If the old and new values are the same, then we don't need to do anything.
-	if (old == obj)
+	// If the old and new values are the same, then we don't need to do anything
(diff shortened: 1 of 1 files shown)


arc.mm
```diff
@@ -838,8 +838,9 @@ static BOOL setObjectHasWeakRefs(id obj)
 	WeakRef *oldRef;
 	id old;
 	loadWeakPointer(addr, &old, &oldRef);
-	// If the old and new values are the same, then we don't need to do anything.
-	if (old == obj)
+	// If the old and new values are the same, then we don't need to do anything
+	// unless we are deleting the weak reference by storing NULL to it.
+	if (old == obj && (obj != NULL || NULL == oldRef))
```
davidchisnall, line 843: Please can you add some brackets around this? I am not sure what the precedenc e is between || and == without looking it up and the next person reading the code might not either.

rfm: Done.

davidchisnall (approved):

PR #336, opened 2025-05-01
Bump robin map version

hmelder:
CMake 4.x.x removed compatibility with versions of CMake older than 3.5 ([Release Notes](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features)).

The Robin Map 1.4.0 bumps `cmake_minimum_required` to address this.


davidchisnall (approved):

davidchisnall:
Thanks. Looks like we also need to bump some CI versions but that can be done separately.

PR #338, opened 2025-07-03
Update encoding2.c - pass pointer to parse_array, not address of pointer

### LINKED ISSUE #337: `objc_alignof_type' returns wrong alignment value for arrays
Greetings. I have observed some incorrect behaviour in `objc_alignof_type' when an array immediately succeeds a structure in a type encoding. A simple example is as follows:

```
#include <objc/encoding.h>
#include <stdio.h>

int
main(void)
{
        const char * ex = "{s=[5i]}";

        printf("Size and alignment: %lu - %lu\n", objc_sizeof_type(ex),
						objc_alignof_type(ex));

        return 0;
}
```

Output:
```
Size and alignment: 20 - 0
```

An alignment of zero is clearly incorrect. I verified that Apple's implementation (through NSGetSizeAndAlignment) returns the correct value:

```
#import <Foundation/Foundation.h>
#include <stdio.h>

int
main(void)
{
	const char * ex = "{s=[5i]}";

	NSUInteger s, a;
	NSGetSizeAndAlignment(ex, &s, &a);
	printf("Size and alignment: %lu - %lu\n", s, a);

	return 0;
}
```

Output:
```
Size and alignment: 20 - 4
```

Providing a string encoding like so does not trigger the issue: `"{s=i[5i]}"`

I narrowed the problem down to the '[' case in `alignof_type`:

```
const char *t = type;
parse_array(&t, (type_parser)alignof_type, &align);
return t;
```

I did all kinds of ugly things like this:

```
size_t e = 0;
parse_array(&t, (type_parser)alignof_type, &e);
*align = max(e, *align);
```

But, it felt like I was merely patching the issue instead of fixing it. I was wondering why e (or align) was being passed by reference if it was already a pointer. It took a while for my brain to ask itself: "wouldn't that be passing the address of the routine's stack-bound argument?" Turns out it was doing exactly that. After removing the ampersand, the alignment value for `"{s=[5i]}"` is correct.

The fix on line 336 of encoding2.c is to remove the `&` on `align`, like so:

`parse_array(&t, (type_parser)alignof_type, align);`

I have not run any further tests whatsoever, but I would be glad to know if anyone else has run into this (admittedly very niche) problem.

cmpwi:
&align results in the value being written to the calling routine's stack-bound argument, not the actual value of align.
(Closes #337)

### DIFF
--- a/encoding2.c
+++ b/encoding2.c
@@ -333,7 +333,7 @@ static const char *alignof_type(const char *type, size_t *align)
 		case '[':
 		{
 			const char *t = type;
-			parse_array(&t, (type_parser)alignof_type, &align);
+			parse_array(&t, (type_parser)alignof_type, align);
 			return t;
 		}
 		case 'b':


encoding2.c
```diff
@@ -333,7 +333,7 @@ static const char *alignof_type(const char *type, size_t *align)
 		case '[':
 		{
 			const char *t = type;
-			parse_array(&t, (type_parser)alignof_type, &align);
+			parse_array(&t, (type_parser)alignof_type, align);
```
davidchisnall, line 336: I think your original proposal from the issue was correct here.  Array alignment should be allowed to increase the alignment required for the type, not decrease it.  See the test case I suggested in the issue.

cmpwi: My apologies! I haven't had much sleep, and I wasn't at my primary computer when I made the PR. I just tried your test case on a FreeBSD VM I spun up with the latest release of libobjc2, but I can't seem to get `"{X=i[2c]}"` to return an alignment of 1, regardless of whatever variation of `parse_array` I try.

I think I'm going to need a little hand-holding with this one. Full disclosure, I just barely managed to wrap my head around this. Which one of my proposals would you like PR'ed?

davidchisnall (approved):

encoding2.c
cmpwi: No worries!

PR #342, opened 2025-07-25
Support 16 KiB Pages on AArch64 Targets

hmelder:
This PR adds support for 16 KiB pages on AArch64. `libobjc2` was also broken on Asahi Linux because of this. Additionally, starting November 1st, 2025, Google requires all applications targeting Android 15+ to support 16 KiB page sizes. 

All hard-coded page sizes in `blocks_to_imp.c` were replaced by one call to `getpagesize()` in the initialization function.
PowerPC CI will probably fail, as the CI is still misconfigured and reports 4 KiB pages.

### DIFF
--- a/block_to_imp.c
+++ b/block_to_imp.c
@@ -206,7 +206,17 @@ static void *trampoline_end_sret;
 PRIVATE void init_trampolines(void)
 {
 	// Retrieve the page size
+	#if defined(__powerpc64__)
+	// For PowerPC we fix the page size to 64KiB.
+	// We therefore effectively support all systems with page size <= 64 KiB.
+	trampoline_page_size = 0x10000;
+	// Check that the pagesize is greater or equal to the smallest size that we
+	// can perform mprotect operations on.
+	assert(pagesize() <= trampoline_page_size);
+	#else
 	trampoline_page_size = pagesize();
+	#endif
+
 	trampoline_region_size = trampoline_page_size * TRAMPOLINE_REGION_PAGES;
 	trampoline_header_per_page = trampoline_page_size / sizeof(struct block_header);
 
--- a/block_to_imp.c
+++ b/block_to_imp.c
@@ -20,11 +20,28 @@
 #include "blocks_runtime.h"
 #include "lock.h"
 #include "visibility.h"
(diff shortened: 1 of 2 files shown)


shims.h
```diff
@@ -0,0 +1,21 @@
...
+#if defined(_WIN32)
+
+/**
+ * getpagesize() returns the current page size
+ */
+int getpagesize(void);
+
+#else // Assume that the platform implements the X/Open System Interface
+
+#include <unistd.h>
```
davidchisnall, line 17: `getpagesize` is not a POSIX function.  I think you want:

```
long getpagesize(void)
{
    return sysconf(_SC_PAGESIZE);
}
```

shims.c
```diff
@@ -0,0 +1,20 @@
+#include <assert.h>
+#include <stdint.h>
+
+#include "shims.h"
+
+#if defined(_WIN32)
+
+#include "safewindows.h"
+
+int getpagesize(void) {
```
davidchisnall, line 10: This API is also not standard, so we could just put this with the platform-specific code at the top of the one file that needs it.

hmelder:
Sorry for the delay. I've moved the definition of `pagesize()` into `block_to_imp.c` and removed the dependency on the non-standard `getpagesize()` function. The changes were tested on Asahi Linux after the rebase.

hmelder:
Oops, I still need to fix the PPC CI.

hmelder:
Turns out, changing the page size of the qemu emulator is apparently not possible, so we either need to disable the CI for powerpc64le, or implement support for 4 KiB pages.

```c
#include <unistd.h>
#include <stdio.h>

int main(void) {
	printf("The page size of the system is %d bytes.\n", getpagesize());
	return 0;
}
```

```sh
qemu-ppc64le: warning: Deprecated page size option cannot change host page size (4096)
The page size of the system is 4096 bytes
```

hmelder:
@davidchisnall the Cirrus CI is not executing due to insufficient compute credits.

davidchisnall:
Cirrus is a bit flaky here, happy to merge if you clean up the history.

PR #356, opened 2025-10-24
Add option to toggle Objective-C exception support

hmelder:
This pull request adds the option `ENABLE_EXCEPTIONS` to the CMake configuration. This option is enabled by default.


hmelder:
EH support for WASM is on its way so there is no need for this PR.

PR #369, opened 2026-01-02
Swap calloc(3) argument order to match parameter names

cmpwi:
Place the count first, and the size of the allocation next.

(Plus, removal of an unused routine in dtable.c, the addition of an `inline` keyword to an always inline routine in arc.mm, and nil-checking simplification in arc.mm. Sorry for not making the commit title more descriptive from the get-go, I didn't think that all commits would be bundled under one pull request!)


davidchisnall (approved):
Thanks!

PR #377, opened 2026-03-25
Implement full support for the LoongArch CPU architecture

basilisk-dev:
Ported objc_msgSend to the LoongArch64 architecture by adding a native assembly implementation. I basically ported the RISC-V implementation to LoongArch64 assembly. You can read more about the LoongArch architecture [here](https://loongfans.cn/en/) if interested.

I ran the relevant tests and they all passed, and I also tested the implementation manually on LoongArch64 to verify it behaves correctly.

Please let me know if you need any additional info.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,6 +109,13 @@ jobs:
             triple: powerpc64le-linux-gnu
             rtld: ld64.so.2
         exclude:
+          # lld versions prior to 18 do not support linking LoongArch in the Ubuntu provided packages.
+          - llvm-version: 16
+            arch:
+              name: loongarch64
+          - llvm-version: 17
+            arch:
+              name: loongarch64
           # FIXME(hugo): Hangs while executing tests.
           - llvm-version: 18
             arch:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -100,6 +100,10 @@ jobs:
             system-processor: riscv64
             triple: riscv64-linux-gnu
             rtld: ld-linux-riscv64-lp64d.so.1
+          - name: loongarch64
+            system-processor: loongarch64
+            triple: loongarch64-linux-gnu
+            rtld: ld-linux-loongarch-lp64d.so.1
           - name: ppc64el
             system-processor: powerpc64le
             triple: powerpc64le-linux-gnu
@@ -118,7 +122,12 @@ jobs:
       - name: Install cross-compile toolchain and QEMU
         run: |
           sudo apt update
-          sudo apt install libstdc++-9-dev-${{ matrix.arch.name }}-cross qemu-user ninja-build
+          if [ "${{ matrix.arch.name }}" = "loongarch64" ]; then
+            # No libstdc++-9 package exists for LoongArch on Ubuntu, so use version 13.
+            sudo apt install libstdc++-13-dev-loong64-cross qemu-user ninja-build
+          else
+            sudo apt install libstdc++-9-dev-${{ matrix.arch.name }}-cross qemu-user ninja-build
+          fi
       - name: Configure CMake
         run: |
           export LDFLAGS="-L/usr/lib/llvm-${{ matrix.llvm-version }}/lib/ -fuse-ld=lld-${{ matrix.llvm-version}} -Wl,--dynamic-linker=/usr/${{ matrix.arch.triple }}/lib/${{ matrix.arch.rtld }},-rpath,/usr/${{ matrix.arch.triple }}/lib"
--- a/block_to_imp.c
+++ b/block_to_imp.c
@@ -118,8 +118,8 @@ struct block_header
 	void *block;
 	void(*fnptr)(void);
 	/*
-	 * On 64-bit platforms, we have 16 bytes for instructions, which ought to
-	 * be enough without padding.
+	 * On most 64-bit platforms, we have 16 bytes for instructions, which ought
+	 * to be enough without padding.
 	 * Note: If we add too much padding, then we waste space but have no other
 	 * ill effects.  If we get this too small, then the assert in
 	 * `init_trampolines` will fire on library load.
@@ -129,10 +129,13 @@ struct block_header
 	 * PAGE_SIZE, so we need to pad block_header to 32 bytes.
 	 * On PowerPC 64-bit where sizeof(void *) = 8 bytes, we
 	 * add 16 bytes of padding.
+	 *
+	 * LoongArch64 needs five 4-byte instructions, so it also requires a
+	 * 32-byte block_header.
 	 */
 #if defined(__i386__) || (defined(__mips__) && !defined(__mips_n64)) || (defined(__powerpc__) && !defined(__powerpc64__))
 	uint64_t padding[3];
-#elif defined(__mips__) || defined(__ARM_ARCH_ISA_A64) || defined(__powerpc64__)
+#elif defined(__mips__) || defined(__ARM_ARCH_ISA_A64) || defined(__powerpc64__) || (defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float))
 	uint64_t padding[2];
 #elif defined(__arm__)
 	uint64_t padding;
@@ -195,6 +198,15 @@ extern char __objc_block_trampoline_16;
 extern char __objc_block_trampoline_end_16;
 extern char __objc_block_trampoline_sret_16;
 extern char __objc_block_trampoline_end_sret_16;
+#elif defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float)
+extern char __objc_block_trampoline_16;
+extern char __objc_block_trampoline_end_16;
+extern char __objc_block_trampoline_sret_16;
+extern char __objc_block_trampoline_end_sret_16;
+extern char __objc_block_trampoline_64;
+extern char __objc_block_trampoline_end_64;
+extern char __objc_block_trampoline_sret_64;
+extern char __objc_block_trampoline_end_sret_64;
 #endif
 
 // Cache the correct trampoline region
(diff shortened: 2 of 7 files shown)


davidchisnall:
Thanks, this looks correct but I’d like to see it running in CI. I believe there is QEMU user mode support for Loongarch, can you add it to the workflow runner?

For this to be useful, we will also need a clang patch that enables the assembly message-dispatch function for message sends.

basilisk-dev:
@davidchisnall I had to make a slight change to allow the implementation to compile on Clang 18 (I used Clang 22 during initial development).

I've added logic to the GitHub Actions workflow to allow the tests to run against LoongArch64 as well. Here is a test run I did: https://github.com/basilisk-dev/libobjc2/actions/runs/23605441514/job/68746795969

I had to exclude LLVM 16 and LLVM 17 on LoongArch64. `ld.lld` in the Ubuntu LLVM 16 and LLVM 17 packages do not support linking on LoongArch64.

A few of the tests fail because there is no LoongArch64 implementation in `block_trampolines.S` but I intend to fix that in a separate PR.
 
> For this to be useful, we will also need a clang patch that enables the assembly message-dispatch function for message sends

Makes sense, I'm not familiar with Clang/LLVM internals so I didn't realize this wasn't already the case.

davidchisnall (approved):
Thanks. Looks like it would be quite easy to tweak to support 32-bit of desirable.

davidchisnall:
CI is failing because of the missing block-to-IMP logic. This is pretty small, but we need to either skip that test on LoongArch or implement it.

davidchisnall:
Assuming the immediate operand on ld.d is large enough on LoongArch, I think you should be able to simply copy [the MIPS code paths here](https://github.com/gnustep/libobjc2/blob/ecef7f1ed1eabf2105995823ec6d6e4693710b0e/block_trampolines.S#L80) and replace `ld` with `ld.d` and `jr` with `jirl` (with a 0 immediate), and define `ARG0` to `ARG2` in exactly the same way.

basilisk-dev:
Hey @davidchisnall, I went through and fixed the remaining feature gaps on LoongArch64. All CICD tests now pass on Loongarch64. The change was pretty minimal, Loongarch supports multiple page sizes like aarch64 so I added some code to account for that. Basically just the same as the MIPS and RISC-V implementations ported to Loongarch assembly.

.github/workflows/main.yml
```diff
@@ -100,11 +100,22 @@ jobs:
...
+          - name: loongarch64
+            system-processor: loongarch64
+            triple: loongarch64-linux-gnu
+            rtld: ld-linux-loongarch-lp64d.so.1
           - name: ppc64el
             system-processor: powerpc64le
             triple: powerpc64le-linux-gnu
             rtld: ld64.so.2
         exclude:
+          # lld versions prior to 18 do not support linking LoongArch in the Ubuntu provided packages.
```
MingcongBai, line 112: Linker relaxation related? Might be good to clarify for the record.

objc/message.h
```diff
@@ -9,6 +9,8 @@
 #if defined(__x86_64) || defined(__i386) || defined(__arm__) ||                \
     defined(__mips_n64) || defined(__mips_n32) ||                              \
     defined(__ARM_ARCH_ISA_A64) ||                                             \
+    (defined(__loongarch__) && defined(__loongarch_lp64) &&                    \
```
MingcongBai, line 12: `__loongarch_lp64` should suffice on its own, cc @xen0n @xry111 right?

block_to_imp.c
```diff
@@ -129,10 +129,13 @@ struct block_header
...
 	 * On PowerPC 64-bit where sizeof(void *) = 8 bytes, we
 	 * add 16 bytes of padding.
+	 *
+	 * LoongArch64 needs five 4-byte instructions, so it also requires a
+	 * 32-byte block_header.
 	 */
 #if defined(__i386__) || (defined(__mips__) && !defined(__mips_n64)) || (defined(__powerpc__) && !defined(__powerpc64__))
 	uint64_t padding[3];
-#elif defined(__mips__) || defined(__ARM_ARCH_ISA_A64) || defined(__powerpc64__)
+#elif defined(__mips__) || defined(__ARM_ARCH_ISA_A64) || defined(__powerpc64__) || (defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float))
```
MingcongBai, line 138: Ditto above regarding arch macro.

block_to_imp.c
```diff
@@ -129,10 +129,13 @@ struct block_header
 	 * PAGE_SIZE, so we need to pad block_header to 32 bytes.
 	 * On PowerPC 64-bit where sizeof(void *) = 8 bytes, we
 	 * add 16 bytes of padding.
+	 *
+	 * LoongArch64 needs five 4-byte instructions, so it also requires a
```
MingcongBai, line 133: Confirming - [*LoongArch Reference Manual - Volume 1: Basic Architecture*](https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html), chapter 1.2 ["Instruction Formats"](https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html#instruction-formats).

block_to_imp.c
```diff
@@ -195,6 +198,15 @@ extern char __objc_block_trampoline_16;
 extern char __objc_block_trampoline_end_16;
 extern char __objc_block_trampoline_sret_16;
 extern char __objc_block_trampoline_end_sret_16;
+#elif defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float)
```
MingcongBai, line 201: Ditto regarding arch macro.

block_to_imp.c
```diff
@@ -241,6 +257,18 @@ PRIVATE void init_trampolines(void)
 		trampoline_start_sret = &__objc_block_trampoline_sret_16;
 		trampoline_end_sret = &__objc_block_trampoline_end_sret_16;
 	} else {
+#elif defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float)
```
MingcongBai, line 260: Ditto.

MingcongBai, line 260: Ditto.

block_trampolines.S
```diff
@@ -156,6 +156,37 @@
 #define SARG0 ARG1
 #define SARG1 ARG2
 
+#elif defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float)
```
MingcongBai, line 159: Ditto.

block_trampolines.S
```diff
@@ -234,7 +265,7 @@ CDECL(__objc_block_trampoline_sret):
 CDECL(__objc_block_trampoline_end_sret):
 
 // Trampoline for 16 KiB page sizes
-#if defined(__ARM_ARCH_ISA_A64)
+#if defined(__ARM_ARCH_ISA_A64) || (defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float))
```
MingcongBai, line 268: Ditto.

block_trampolines.S
```diff
@@ -247,6 +278,20 @@ CDECL(__objc_block_trampoline_sret_16):
 CDECL(__objc_block_trampoline_end_sret_16):
 #endif
 
+// Trampoline for 64 KiB page sizes
+#if defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float)
```
MingcongBai, line 282: Ditto.

objc_msgSend.S
```diff
@@ -10,6 +10,8 @@
 #include "objc_msgSend.aarch64.S"
 #elif defined(__riscv) && (__riscv_xlen == 64) && defined(__riscv_float_abi_double)
 #include "objc_msgSend.riscv64.S"
+#elif defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float)
```
MingcongBai, line 13: Ditto.

MingcongBai (changes requested):
Preliminary review from me, I might have missed/misinterpreted something.

@xen0n @xry111 @ziyao233, etc. Would really appreciate your input.

block_trampolines.S
```diff
@@ -156,6 +156,37 @@
 #define SARG0 ARG1
 #define SARG1 ARG2
 
+#elif defined(__loongarch__) && defined(__loongarch_lp64) && defined(__loongarch_double_float)
+////////////////////////////////////////////////////////////////////////////////
+// LoongArch64 trampoline
+////////////////////////////////////////////////////////////////////////////////
+.macro trampoline arg0, arg1
+	pcaddi   $t0, -1024
```
davidchisnall, line 164: The LoongArch ABI doesn’t do the MIPS trick of requiring a specific register for jalr so you know the PC on entry?

block_trampolines.S
```diff
@@ -156,6 +156,37 @@
...
+	jr       $t0
+.endm
+.macro trampoline_16 arg0, arg1
+	pcaddi   $t0, -4096
+	move     \arg1, \arg0
+	ld.d     \arg0, $t0, 0
+	ld.d     $t0, $t0, 8
+	jr       $t0
+.endm
+.macro trampoline_64 arg0, arg1
```
davidchisnall, line 177: These trampolines appear to differ only by the displacement from the start, can’t that be just another macro argument?

block_trampolines.S
xry111: No it doesn't, the MIPS way is mostly a hack for lacking the instructions like pcaddi.

objc/message.h
ziyao233: `__loongarch__` should be redundant, but I think we should guard against `__loongarch_double_float` since floating registers are spilled/restored in the code as well.

PR #380, opened 2026-04-17
Update compile options to enable static linking to work

ingresso-pete:
This makes the static options the same as the dynamic, which was causing the linking to fail. As this requires robin-map, and this is installed under /usr/local on BSD and similar systems, then we also add /usr/local/include to the header search path if it exists.

All the tests pass, and this works perfectly with my test code too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,9 @@ add_compile_definitions($<$<BOOL:${STRICT_APPLE_COMPATIBILITY}>:STRICT_APPLE_COM
 
 configure_file(objc/objc-config.h.in objc/objc-config.h @ONLY)
 include_directories("${PROJECT_BINARY_DIR}/objc/")
+if (IS_DIRECTORY /usr/local/include)                                           
+	include_directories("/usr/local/include")
+endif()
 
 if (EMBEDDED_BLOCKS_RUNTIME)
 	list(APPEND libobjc_ASM_SRCS block_trampolines.S)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -155,6 +155,9 @@ add_compile_definitions($<$<BOOL:${STRICT_APPLE_COMPATIBILITY}>:STRICT_APPLE_COM
 
 configure_file(objc/objc-config.h.in objc/objc-config.h @ONLY)
 include_directories("${PROJECT_BINARY_DIR}/objc/")
+if (IS_DIRECTORY /usr/local/include)                                           
+	include_directories("/usr/local/include")
+endif()
```
davidchisnall: I'm not sure why this is needed, we should be fetching the robin map if it isn't installed, or finding it otherwise.  I don't like accidentally depending on ambient things if we can avoid it.

davidchisnall (commented):
Please can you also add static to at least some of the test matrix so that we don't break it in the future?

CMakeLists.txt
ingresso-pete: yes, it puzzled me, because even though it includes the header, you dont need to link the robin library to use the result. will take a look at this.

will also look at the tests, [ I assumed that they would run automatically if I built it static ]

ingresso-pete: [ oh, the code is all in the header.... ]

ingresso-pete: For the tests, if you build the static library, it now compiles a static version of each of the non-legacy tests, and statically links the whole thing. Those get run alongside the shared library tests, and they all pass.

[ I am not great at CMake, so I have to admit I used Claude for the actual changes to make this happen. I understand what it has done, and it seems reasonable to me, but you may want to eyeball that last commit a bit more closely. ]

davidchisnall (approved):

CMakeLists.txt
davidchisnall: Yes, CMake is a bit weird, ‘link library’ really means ‘import everything this target exposes’, which includes adding include paths it exports.

PR #382, opened 2026-04-19
Fix conflicting checks for stop point in arc.mm:emptyPool()

cmpwi:
The code used NULL before, so I decided to stick with that.

### DIFF
--- a/arc.mm
+++ b/arc.mm
@@ -155,57 +155,71 @@ static inline void release(id obj);
 
 /**
  * Empties objects from the autorelease pool, stating at the head of the list
- * specified by pool and continuing until it reaches the stop point.  If the stop point is NULL then 
+ * specified by pool and continuing until it reaches the stop point.  If the stop
+ * point is NULL then all pools are cleared.
  */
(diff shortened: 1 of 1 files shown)


davidchisnall:
Looks like the tests that cover this code are now all failing.

arc.mm
```diff
@@ -155,57 +155,71 @@ - (void)release;
...
+	void *oldPool;
+
+	/* Are we clearing up to a given object? */
+	if (stopAt != NULL)
 	{
 		stopPool = tls->pool;
-		while (1)
+
+		/* Find pool in which object to stop at is located. */
+		while (stopPool != NULL)
```
davidchisnall, line 173: This condition is now dead. We leave the loop via either break or return, never by this condition failing.

cmpwi:
Oopsie! The insert point shouldn't go behind the pool if new ones are allocated. Should be all good now!

davidchisnall (approved):

arc.mm
cmpwi: I'm not sure what you mean. The return is after the loop, yes?

PR #384, opened 2026-04-22
Replace hard-coded four spaces with tab in objc_alloc()

### LINKED ISSUE #387: `class_getMethodImplementation()` calls forwarding hook with nil receiver
`class_getMethodImplementation()` calls `__objc_msg_forward2()` with a nil receiver when an IMP on the class cannot be found. Is this correct? This behaviour currently makes it impossible for the hook to perform any kind of forwarding so that `class_getMethodImplementation()` returns the final IMP.

Apple's implementation of `class_getMethodImplementation()` returns the forwarding hook in this case, and its documentation states the following:

> Returns the function pointer that would be called if a particular message were sent to an instance of a class.

This is much better, but in my understanding, the description isn't 100% true without extra caller intervention; if the returned IMP is the forwarding hook, you'd have to call the hook to get the real IMP or crash with an unknown selector, which correctly mirrors `objc_msgSend()`'s behaviour. With the current behaviour, given a hook that returns nil when called with a nil receiver and a nonnull SEL, an implementation of, say, `-performSelector:` would have to look like this:

```
- (id)performSelector:(SEL)aSelector
{
	IMP i;

	if (aSelector == nullptr)
		[self doesNotRecognizeSelector:aSelector];

	i = class_getMethodImplementation(object_getClass(self), aSelector);

	/*
	 * __objc_msg_forward2(nil, aSelector) was called; it returns nil to inform -performSelector: that the method
	 * wasn't found. We can't perform any forwarding, but we can fail successfully with a proper exception message
	 * showing self's class name instead of throwing a generic exception in the forwarding hook.
	 */
	if (i == nullptr)
		[self doesNotRecognizeSelector:aSelector];

	/* Call the real method. */
	return ((id (*)(id, SEL))i)(self, aSelector);
}
```

With Apple's behaviour, an implementation of `-performSelector:` could look like this:

```
- (id)performSelector:(SEL)aSelector
{
	IMP i;

	if (aSelector == nullptr)
		[self doesNotRecognizeSelector:aSelector];

	i = class_getMethodImplementation(object_getClass(self), aSelector);

	/* Method not found, forwarding hook returned. Call it and get the real method to invoke. */
	if (i == __objc_msg_forward2)
		i = ((IMP (*)(id, SEL))i)(self, aSelector);

	/* Call the real method. */
	return ((id (*)(id, SEL))i)(self, aSelector);
}
```

This would be behaviourally equivalent to a much simpler `objc_msgSend()` approach:

```
- (id)performSelector:(SEL)aSelector
{
	if (aSelector == nullptr)
		[self doesNotRecognizeSelector:aSelector];

	return ((id (*)(id, SEL))objc_msgSend)(self, aSelector);
}
```

Which, in turn, is also behaviourally equivalent to GNUstep's approach of calling `objc_msg_lookup()`. Currently, the only correct way to implement `-performSelector:` without depending on the hook's behaviour *and* have forwarding support is to go through the `objc_msgSend()` or `objc_msg_lookup()`.

I do admit that using `-performSelector:` as an example is a bit contrived, since it should probably only be implemented with `objc_msgSend()` or `objc_msg_lookup()`. But, it was the simplest example I could come up with to illustrate my point.

I think that returning the forwarding hook from `class_getMethodImplementation()` is worth it despite probably being a non-backwards-compatible behavioural change. Ignoring the whole "feature parity with Apple" argument, clear benefits include:

- Allowing for callers to perform forwarding manually by feeding the forwarding hook a valid, nonnull object instead of only being able to crash.
- Making the hook less fragile; it should only return a valid final IMP after any potential forwarding (as per its header; its return value is not marked `_Nullable` and no comment states that it could or should return nullptr), or crash on a truly unimplemented selector.

Thoughts? Opinions?

cmpwi:
Super inconsequential nitpick, but every other line in `objc_alloc()` uses tabs, except for that last line (this is probably only visible if you have tabs set to 8 spaces, but invisible in the editor view if its set to 4 spaces.)

(EDIT: looks like my `class_getMethodImplementation()` PR got rolled into this prior commit. Closes #387.)


davidchisnall (approved):

davidchisnall:
Can you rebase this to get rid of the merge?

PR #385, opened 2026-05-18
Fix atomic property setters held spinlock during release causing deadlocks

johnathan-becker:
Atomic property setters were releasing the old value while still holding the spinlock. Since release can invoke dealloc, which may access other atomic properties, this deadlocked on re-entry. Also fixed DoubleLockGuard swapping constructor parameters instead of member variables, which left members in unsorted order and triggered the destructor assert.


davidchisnall (approved):

johnathan-becker:
@davidchisnall I believe a workflow needs started here.

PR #386, opened 2026-05-18
objc_msgSend: powerpc64* support

pkubaj:
Works on both little- and big-endian.

### DIFF
--- a/objc_msgSend.powerpc64.S
+++ b/objc_msgSend.powerpc64.S
@@ -1,0 +1,298 @@
+/* objc_msgSend.powerpc64.S - PowerPC64 ELFv2 fast-path dispatch for GNUstep libobjc2.
+ *
+ * Included from objc_msgSend.S after common.S and asmconstants.h.
+ * Works for both ppc64le (LE) and ppc64 BE with the ELFv2 ABI.
+ *
+ * Dispatch overview:
+ *   Fast path (no frame): nil-check → small-obj-check → isa → dtable lookup → IMP → bctr
+ *   Slow path (frame + spill): for nil slots (cache miss), calls slowMsgLookup
+ *
+ * ELFv2 frame layout used on the slow path (FRAME_SIZE = 208):
+ *   +0   back chain   +8  CR save   +24  TOC (r2)
+ *   +32  r3  +40  r4  +48  r5  +56  r6  +64  r7  +72  r8  +80  r9  +88  r10
+ *   +96  f1  ...  +192  f13
+ *   LR saved at (FRAME_SIZE+16)(r1) = old_sp+16 per ELFv2 convention.
+ */
+
+#define FRAME_SIZE 208
+
+.globl CDECL(objc_msgSend)
+TYPE_DIRECTIVE(CDECL(objc_msgSend), %function)
+.globl CDECL(objc_msgSend_fpret)
+TYPE_DIRECTIVE(CDECL(objc_msgSend_fpret), %function)
+
+CDECL(objc_msgSend_fpret):
+CDECL(objc_msgSend):
+.Lgep_objc_msgSend:
+	.cfi_startproc
+	addis	2, 12, .TOC.-.Lgep_objc_msgSend@ha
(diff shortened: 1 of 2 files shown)


davidchisnall:
This looks technically correct, but seems to be somewhat missing the point of the assembly implementations. This always hits the slow path and will be slower than the portable mechanism.

objc_msgSend.powerpc64.S
```diff
@@ -0,0 +1,226 @@
...
+	addi  2, 2, .TOC.-.Lgep_objc_msgSend@l
+.localentry CDECL(objc_msgSend), .-.Lgep_objc_msgSend
+
+	/* Nil check: r3 = self */
+	cmpdi	cr0, 3, 0
+	beq-	cr0, .Lnil_msgSend
+
+	/* Prologue: save LR, create frame, save TOC */
+	mflr	0
+	stdu	1, -FRAME_SIZE(1)
```
hmelder, line 100: I think you want a `cfi_def_cfa_offset` here.

objc_msgSend.powerpc64.S
```diff
@@ -0,0 +1,226 @@
...
+	lfd	7,  144(1)
+	lfd	8,  152(1)
+	lfd	9,  160(1)
+	lfd	10, 168(1)
+	lfd	11, 176(1)
+	lfd	12, 184(1)
+	lfd	13, 192(1)
+
+	/* Tear down frame and restore LR */
+	addi	1, 1, FRAME_SIZE
```
hmelder, line 155: And here a `cfi_adjust_cfa_offset` again

objc_msgSend.powerpc64.S
```diff
@@ -0,0 +1,226 @@
...
+	stfd	11, 176(1)
+	stfd	12, 184(1)
+	stfd	13, 192(1)
+
+	/* Call objc_msg_lookup(self=r3, sel=r4) - args already in place */
+	bl	CDECL(objc_msg_lookup)
+	nop
+
+	/* Move returned IMP to r12 (required for ELFv2 indirect calls) */
+	mr	12, 3
```
hmelder, line 133: Is there a reason for using register indices instead of `%r<NUMBER>` and `%f<NUMBER>`? It makes it cumbersome to read.

objc_msgSend.powerpc64.S
```diff
@@ -0,0 +1,226 @@
...
+	li	3, 0
+	li	4, 0
+	blr
+	.cfi_endproc
+
+
+.globl CDECL(objc_msgSend_stret)
+TYPE_DIRECTIVE(CDECL(objc_msgSend_stret), %function)
+
+/* For stret methods: r3=hidden struct ptr, r4=self, r5=sel */
```
hmelder: You can use a macro and save the duplication here. Checkout my draft `objc_msgSend.powerpc.S` here: https://github.com/gnustep/libobjc2/blob/ffa2b4fd3a96fbaa5440d3a53a66812d4b38de6f/objc_msgSend.powerpc.S#L177

hmelder:
I agree with David that there is no point in having this assembly trampoline if you always take the slow path. Maybe base it around my already existing draft. The draft is basically a translation of the RISC-V trampoline.

https://github.com/gnustep/libobjc2/blob/ffa2b4fd3a96fbaa5440d3a53a66812d4b38de6f/objc_msgSend.powerpc.S#L2

hmelder:
Also enabling this requires an additional patch in `CGObjCGNU.cpp` in Clang codegen to actually emit calls to `objc_msgSend` and friends.

pkubaj:
Updated. A disclosure: I'm not an assembly dev and this was coded with Claude Opus 4.7. The full test suite doesn't pass, but the failure count is the same pre-patch and post-patch. libobjc2 doesn't seem to have an anti-AI policy, so I hope it's fine.

davidchisnall:
We do not accept LLM contributions at this time.  I didn't feel the need to write it down, because assuming that people actually wrote and understood their PRs has been the societal norm for decades.

PR #389, opened 2026-06-28
spaces->tabs

cmpwi:
Consistency.

### DIFF
--- a/fast_paths.m
+++ b/fast_paths.m
@@ -25,7 +25,7 @@ objc_alloc(Class cls)
 	{
 		return class_createInstance(cls, 0);
 	}
-    return [cls alloc];
+	return [cls alloc];
 }
 
 /**


davidchisnall (approved):

davidchisnall:
Recent versions of clang-format can now (finally!) handle the style used in this project (tabs for indentation, spaces for alignment), so I should probably start using it.

The config in CHERIoT RTOS is close to what I use here, but doesn’t have the Objective-C bits set. If you feel like adding them and adding a format check CI job, that would be very helpful.

cmpwi:
You're referring to this file: [https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/.clang-format](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/.clang-format), right?

I've never actually used clang-format before, but I'm more than willing to learn. I'll take a look after the sun goes up in my timezone (what is sleep?) and try some experiments on my own code so I don't turn yours into mojibake. After that, I'll figure out how to wire it into the CI system.

davidchisnall:
> You're referring to this file: https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/.clang-format, right?

Yup.

> I've never actually used clang-format before, but I'm more than willing to learn. I'll take a look after the sun goes up in my timezone (what is sleep?) and try some experiments on my own code so I don't turn yours into mojibake. After that, I'll figure out how to wire it into the CI system.

The cheriot-rtos repo also has a CI job that does the formatting, so we can probably largely copy that one.  It basically runs clang-format, then runs git diff and, if there is a diff, fails and reports the error.  

Running it on (a copy of) the code in this repo is a good idea because git diff will show you how close the output is to the existing style.  Hopefully you can get it to the point where the changes are all places where I haven't been consistent in applying the style here.

Issue #398, opened 2026-07-14
Is multicore ARC scaling in scope for libobjc2? some retain/release and weak-reference contention findings


DTW-Thalion:
We run a number of GNUstep-based services at Thalion on a high-throughput server farm, and we have been trying to squeeze more performance out of the runtime under that kind of load (many cores, heavily multithreaded, long-lived processes). Before we take this any further or propose anything concrete, I wanted to check with you whether this is even a direction the project cares about, because how much of it is worth doing depends a lot on what a typical libobjc2 workload actually looks like.

My impression is that a large share of GNUstep use is GUI or otherwise single-threaded, where the runtime already performs very well. The things we are running into only really show up under sustained multicore contention. So the first question is: do you consider the many-core, heavily-threaded case to be in scope for libobjc2, or are most users single-threaded enough that improving multicore scaling would be effort spent on a case few people would benefit from? If the answer is "most workloads are single-threaded", so be it, what follows may not be worht inclusion in the master repo.

To make the question concrete, here is what we found: We profiled the hot paths under 1/2/4-thread scaling. The first thing worth saying is that dispatch is already excellent and we do not think it should be touched: it is lock-free, it scales flat across cores, and polymorphic dispatch measures the same as monomorphic (about 3.0 vs 3.2 ns), so the trie is genuinely O(1) with no cache-thrash penalty. Swapping in objc4-style per-class hash caches would be a lateral move at best and would give up the never-miss property. 

The operations that scale badly are all in ARC and the metadata locks, not dispatch.

The one finding that helps even single-threaded callers is the retain/release atomics. A retain/release pair is about 31 ns, and an isolated micro-test (no Objective-C, just the atomic pattern) shows that nearly all of that is the atomic pattern itself:

```
libobjc2-style pair  (fetch_add(p,0) seed + CAS, seq_cst)   27.6 ns
relaxed-load seed    + acq/rel CAS                          14.4 ns
plain fetch_add/fetch_sub (relaxed)                          8.1 ns
```

[...]

PR #399, opened 2026-07-14
Replace __sync reference-count atomics with std::atomic

DTW-Thalion:
Per #398.  The reference-count fast paths in arc.mm seeded their CAS loops with `__sync_fetch_and_add(refCount, 0)`, a full read-modify-write used only to read the count word, and updated it with sequentially consistent `__sync_val_compare_and_swap`.  The seed only needs a plain load, and the exchanges do not need seq_cst.

This views the count word through `std::atomic<uintptr_t>` and operates on it directly.  A relaxed load seeds each loop; `compare_exchange_weak` is acquire-release on the retain and weak-flag paths and release on the decrement; the final release takes an acquire fence before running `-dealloc`.  No functional or ABI change.

Isolating just the atomic pattern with no Objective-C involved, on a 32-core x86-64 host with clang 18.1.3:

```
__sync seed + CAS, seq_cst        27.6 ns
relaxed load seed + acq/rel CAS   14.4 ns
```

and a retain/release pair on a real object drops from about 31 ns to about 16 ns.

All 194 tests pass, along with an 8-thread retain/release balance stress where the count returns to its exact starting value and an 8-thread weak load/store/dealloc race stress.

### DIFF
--- a/arc.mm
+++ b/arc.mm
@@ -6,6 +6,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <assert.h>
+#include <atomic>
 #include <vector>
 #include <tsl/robin_map.h>
 #import "lock.h"
@@ -270,19 +271,18 @@ static const size_t refcount_max = refcount_mask - 1;
 
 extern "C" OBJC_PUBLIC size_t object_getRetainCount_np(id obj)
 {
-	uintptr_t *refCount = ((uintptr_t*)obj) - 1;
-	uintptr_t refCountVal = __sync_fetch_and_add(refCount, 0);
+	auto *refCount = reinterpret_cast<std::atomic<uintptr_t>*>(obj) - 1;
+	uintptr_t refCountVal = refCount->load(std::memory_order_relaxed);
 	size_t realCount = refCountVal & refcount_mask;
 	return realCount == refcount_mask ? 0 : realCount + 1;
 }
 
 static id retain_fast(id obj, BOOL isWeak)
 {
-	uintptr_t *refCount = ((uintptr_t*)obj) - 1;
-	uintptr_t refCountVal = __sync_fetch_and_add(refCount, 0);
-	uintptr_t newVal = refCountVal;
-	do {
-		refCountVal = newVal;
+	auto *refCount = reinterpret_cast<std::atomic<uintptr_t>*>(obj) - 1;
+	uintptr_t refCountVal = refCount->load(std::memory_order_relaxed);
+	for (;;)
+	{
 		size_t realCount = refCountVal & refcount_mask;
 		// If this object's reference count is already less than 0, then
 		// this is a spurious retain.  This can happen when one thread is
@@ -310,9 +310,16 @@ static id retain_fast(id obj, BOOL isWeak)
 		realCount++;
 		realCount |= refCountVal & weak_mask;
 		uintptr_t updated = (uintptr_t)realCount;
-		newVal = __sync_val_compare_and_swap(refCount, refCountVal, updated);
-	} while (newVal != refCountVal);
-	return obj;
+		// Acquire/release on the exchange so reference-count updates are
+		// ordered against each other on weakly-ordered targets.  On a failed
+		// exchange refCountVal is refreshed with the current value.
+		if (refCount->compare_exchange_weak(refCountVal, updated,
+		                                    std::memory_order_acq_rel,
+		                                    std::memory_order_acquire))
+		{
+			return obj;
+		}
+	}
 }
 
 extern "C" OBJC_PUBLIC id objc_retain_fast_np(id obj)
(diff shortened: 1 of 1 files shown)


arc.mm
```diff
@@ -271,18 +271,17 @@ static TLS_CALLBACK(cleanupPools)(struct arc_tls* tls)
 extern "C" OBJC_PUBLIC size_t object_getRetainCount_np(id obj)
 {
 	uintptr_t *refCount = ((uintptr_t*)obj) - 1;
-	uintptr_t refCountVal = __sync_fetch_and_add(refCount, 0);
```
davidchisnall, line 274: Can we use `std::atomic` instead of builtins?

DTW-Thalion: Done. Now uses `std::atomic_ref<uintptr_t>` over the count word instead of the builtins, keeping the same orderings: a relaxed load seeds each loop, `compare_exchange_weak` is relaxed on the retain and weak-flag paths and release on the decrement, and there's an acquire fence before `-dealloc`. All 194 tests still pass, along with the retain/release balance and weak-race stresses. Should mention that this seems to work with C++20 but may not work on earlier versions.

The other option was reinterpret_cast<std::atomic<uintptr_t>(refcount) which has the upside of working on any C++11+ standard library. We can revisit that if you prefer. The downside is this is technically UB - accessing an object that wasn't created as a std::atomic through an atomic pointer.  So I went with the standard way at the expense of a libc++ >= 19.

davidchisnall: I was more thinking of chaning the previous line to:

```c++
	auto *refCount = reinterpret_cast<std::atomic<uintptr_t>*>(obj) - 1;
```

DTW-Thalion: Done, that's cleaner. Switched all five sites to `reinterpret_cast<std::atomic<uintptr_t>*>(obj) - 1` and operate on it directly. Same orderings; all 194 tests plus the retain/release balance and weak-race stresses still pass.

arc.mm
```diff
@@ -310,9 +310,16 @@ static id retain_fast(id obj, BOOL isWeak)
 		realCount++;
 		realCount |= refCountVal & weak_mask;
 		uintptr_t updated = (uintptr_t)realCount;
-		newVal = __sync_val_compare_and_swap(refCount, refCountVal, updated);
-	} while (newVal != refCountVal);
-	return obj;
+		// A retain needs only atomicity of the increment, not ordering
+		// against other memory accesses, so relaxed ordering suffices.  On
+		// a failed exchange refCountVal is refreshed with the current value.
```
davidchisnall: I don't believe this is correct.  Reference count manipulations need to be ordered.  I think the CAS needs to be `rel_acq`.  A relaxed-consistency load in a loop is allowed to be optimised to a single load, because there's no happens-before relationship established with other threads that do the store.  The requirement for a reference count store is that it must establish a happens-before edge to other threads doing a reference-count load.

arc.mm
```diff
@@ -826,8 +843,17 @@ static BOOL setObjectHasWeakRefs(id obj)
 			// reference and so it shouldn't be possible to deallocate it
 			// while we're assigning it.
 			uintptr_t updated = ((uintptr_t)realCount | weak_mask);
-			newVal = __sync_val_compare_and_swap(refCount, refCountVal, updated);
-		} while (newVal != refCountVal);
+			// Relaxed suffices: the flag lives in the reference-count word and
+			// the exchange is itself a read-modify-write, so it observes the
+			// word's modification order.  The weak-ref lock, held here, orders
+			// the weak-table entry we publish next.
+			if (refCount->compare_exchange_weak(refCountVal, updated,
```
davidchisnall, line 849: As above.

arc.mm
```diff
@@ -907,8 +933,8 @@ static BOOL setObjectHasWeakRefs(id obj)
 	if (objc_test_class_flag(classForObject(obj), objc_class_flag_fast_arc))
 	{
 		// Don't proceed if the object isn't deallocating.
-		uintptr_t *refCount = ((uintptr_t*)obj) - 1;
-		uintptr_t refCountVal = __sync_fetch_and_add(refCount, 0);
+		auto *refCount = reinterpret_cast<std::atomic<uintptr_t>*>(obj) - 1;
+		uintptr_t refCountVal = refCount->load(std::memory_order_relaxed);
```
davidchisnall, line 936: This one is okay, but only because we're relying on the caller already having done the synchronisation to ensure that this cannot be concurrently modified.

arc.mm
DTW-Thalion: I had to think about this a bit, but I think I have no problem making the change you are suggesting, though it took me a second to get there. :)

I'd gone relaxed on the increment for the usual shared_ptr reason: relaxed retain, release on the decrement, acquire fence before `-dealloc`. A retain only ever happens off a reference you already hold, so it carries no data of its own and the synchronisation all sits on the release side, which this keeps. But acq_rel is a safe superset and effectively free here. On x86 the `lock cmpxchg` is a full barrier regardless of the order argument, so it only changes codegen on weakly-ordered targets like aarch64, and the speedup in the numbers came from dropping the seq_cst `fetch_add(refCount, 0)` seed rather than from relaxed-vs-acq_rel on the exchange itself. So there's no real cost to tightening it, and it removes any doubt.

I'll switch the retain and weak-flag CAS to acq_rel and repush.

davidchisnall (approved):

triplef:
Nice, thank you for this optimization @DTW-Thalion!

PR #401, opened 2026-07-23
ARC: shard the weak-reference table

DTW-Thalion:
The weak-reference table took a single global lock, so weak operations on unrelated objects serialised on one mutex and scaled negatively with core count.

The striping now lives inside the table type rather than at its call sites: WeakRefTable holds an array of independently locked stripes, picks one from the object address, and keeps the operations it offered before, so nothing outside it knows the table is striped at all. The stripe count is a build flag or the OBJC_WEAK_SHARD_COUNT environment variable.

Three things the striping needed in its own right: the stripe a block belongs to is fixed when the block is created, which closes a lock-selection race that showed up as a use-after-free under AddressSanitizer with more than one stripe; a block's weak slot is accessed atomically, so the striping also holds on weakly ordered targets, checked under ThreadSanitizer; and a block is recycled within its own stripe.

### DIFF
--- a/arc.mm
+++ b/arc.mm
@@ -690,12 +690,21 @@ static int weakref_class;
 
 namespace {
 
+// Weak-reference control block.  `shardIndex` is the owning stripe: set once,
+// never changed, and blocks are recycled (never freed), so a slot-first
+// operation can read it to pick the stripe without a lock.
 struct WeakRef
 {
-	void *isa = &weakref_class;
+	void *isa;
 	id obj = nullptr;
 	size_t weak_count = 1;
-	WeakRef(id o) : obj(o) {}
+	size_t shardIndex;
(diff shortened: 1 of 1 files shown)


davidchisnall:
This looks more invasive than it should be. The original refactoring to make the type used for a weak ref table distinct was intended to make it easy to provide a replacement that did the sharding internally.

DTW-Thalion:
The commits on this branch reached master through #403, which was stacked on
this branch and was merged on 2 August. Master carries the WeakRefTable
encapsulation asked for above: `class WeakRefTable` and
`using weak_table_t = WeakRefTable<OBJC_WEAK_SHARD_COUNT>` are both in arc.mm.

`git cherry master arc-weak-table-sharding` reports both commits as already
upstream, and rebasing this branch onto master leaves no commits and no diff,
so there is nothing left here to review. Closing as merged.

PR #402, opened 2026-07-23
ARC: retain and release with fetch_add instead of a CAS loop

DTW-Thalion:
The strong retain and release fast paths ran a compare-exchange loop that retried on every lost race. A strong retain already holds a live reference, so the object cannot be at the deallocating sentinel and the increment can be a single fetch_add. Release becomes a single fetch_sub, with the last-reference and saturation cases handled after the fact. The weak-to-strong retain keeps its compare-exchange loop, since it can race a concurrent final release and has to test and increment together.

The count manipulation is behind a RefCount class rather than open-coded masks at each call site, obtained with RefCount::fromObject(obj). The weak flag, the guard bit that keeps a saturating increment from carrying into it, the deallocating sentinel and the saturation rule are all stated once, inside that type. Keeping it behind one type is also what would let the count move into spare isa bits, with overflow spilling to a side table, without touching any caller.

Uncontended retain and release drop from 16.1 to 11.3 ns on a 32-core machine with 24 threads active, and a single shared object under 24 threads from 2143 to 1124 ns.

### DIFF
--- a/Test/FastRefCount.m
+++ b/Test/FastRefCount.m
@@ -85,7 +85,8 @@ int main()
 
 const long refcount_shift = 1;
 const size_t weak_mask = ((size_t)1)<<((sizeof(size_t)*8)-refcount_shift);
-const size_t refcount_mask = ~weak_mask;
+const size_t guard_mask = weak_mask >> 1;
+const size_t refcount_mask = ~(weak_mask | guard_mask);
 const size_t refcount_max = refcount_mask - 1;
 
 size_t get_refcount(id obj)
--- a/arc.mm
+++ b/arc.mm
@@ -256,70 +256,215 @@ static IMP AutoreleaseAdd;
 
 static BOOL useARCAutoreleasePool;
 
-static const long refcount_shift = 1;
+namespace {
 /**
- * We use the top bit of the reference count to indicate whether an object has
- * ever had a weak reference taken.  This lets us avoid acquiring the weak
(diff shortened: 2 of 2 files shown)


davidchisnall:
I think this approach is good, but these paths are starting to get quite cluttered. I’d like to have an explicit refcount class that has a from-object static factory method so that the refcount manipulation can be pulled out. That would also let us do the Apple-style refcount-stored-in-the-isa-pointer-with-overflow-in-a-table model.

DTW-Thalion:
The commits on this branch reached master through #403, which was stacked on
this branch and was merged on 2 August. arc.mm on this branch is identical to
arc.mm on master, `git cherry master arc-retain-fetch-add` reports both commits
as already upstream, and rebasing onto master leaves no commits and no diff.
Closing as merged.

The explicit refcount class with a from-object factory method suggested above
is not part of what landed, so that one is still open as a change to make.

DTW-Thalion:
Correction to the note above: the refcount class is in master. arc.mm has
`class RefCount` with `static RefCount fromObject(id obj)`, holding the bit
layout and the atomic manipulation, and the retain, release and weak entry
points go through it as `retainCount`, `increment`, `decrement`,
`markWeaklyReferenced` and `isDeallocating`. Its comment records that the
layout is kept behind the one type so the count can later move into spare isa
bits with overflow in a side table without touching the callers.

The statement that it was not part of what landed was wrong.

PR #403, opened 2026-07-23
ARC: optional cache-line isolation of the reference count

DTW-Thalion:
Reference counts of distinct objects that share a cache line ping-pong between cores when those objects are retained and released concurrently, so unrelated objects contend even though they share no reference.

The reference-count word sits immediately before the object, so aligning each allocation to a cache line puts each object's count on a line of its own. Distinct-object retain and release at four threads goes from 127 to 24 ns. The cost is memory: the smallest instances round up, roughly doubling their footprint, so the alignment is a build setting, OBJC_ALLOC_ALIGN, and a build can trade the scaling back for the memory.

The larger alignment is requested only where it is actually larger than the alignment the plain allocator already gives, tested against alignof(max_align_t) rather than against a written-out number, and a value that is not a power of two is rejected at compile time rather than failing every allocation at run time.

On the question of the allocator: snmalloc does not remove the cliff. Allocated together, it packs 16-byte instances onto shared lines much as glibc does, and is worse at high thread counts; its size-class rounding does not spread them apart.

### DIFF
--- a/gc_none.c
+++ b/gc_none.c
@@ -5,18 +5,57 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#include <stddef.h>
+
+// Alignment of object allocations.  The reference-count word precedes the
+// object and thus sits at the head of the block, so aligning the block to a
+// cache line puts each object's reference count on its own line: two distinct
+// objects are then >= one line apart and never share a line for their counts.
+// That eliminates the false-sharing cliff where independent threads
+// retaining/releasing adjacent small objects ping-pong a shared line
+// (measured: distinct-object retain/release at 4 threads 127ns -> 24ns).
+//
+// This is the default.  The cost is memory: cache-line alignment rounds every
+// small allocation up to the alignment, which roughly DOUBLES the footprint of
+// the smallest objects (measured: 32 -> 64 bytes RSS for a 16-byte instance).
+// Build with -DOBJC_ALLOC_ALIGN=16 (32 on Windows, the minimum vector ivars
+// need) to trade the multicore retain/release scaling back for that memory.
+#ifndef OBJC_ALLOC_ALIGN
+#  define OBJC_ALLOC_ALIGN 64
+#endif
+// The alignment reaches posix_memalign (and _aligned_malloc on Windows), which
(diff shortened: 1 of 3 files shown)


gc_none.c
```diff
@@ -6,17 +6,47 @@
...
 
+// Alignment of object allocations.  The reference-count word precedes the
+// object and thus sits at the head of the block, so aligning the block to a
+// cache line puts each object's reference count on its own line: two distinct
+// objects are then >= one line apart and never share a line for their counts.
+// That eliminates the false-sharing cliff where independent threads
+// retaining/releasing adjacent small objects ping-pong a shared line
+// (measured: distinct-object retain/release at 4 threads 127ns -> 24ns).
+//
+// The cost is memory: cache-line alignment rounds every small allocation up to
```
davidchisnall: Note that this is specific to the malloc implementation used. Handling aligned allocations of things that are not a multiple of the alignment size is painful, so snmalloc rounds up the size.

I am also curious how much of the false-sharing overhead is an artefact of the allocation policy. Do you know how snmalloc affects your workload with and without this relative to the baseline?

DTW-Thalion: Good question - so yesterday I wrote a test to try to answer that and I sent you an e-mail on the results. Short answer:  snmalloc does not remove the cliff. When the objects are allocated together it packs the 16-byte instances onto shared cache lines the same as glibc, and is in fact worse at high thread counts. Its size-class rounding does not spread these apart.

gc_none.c
```diff
@@ -6,17 +6,41 @@
...
 	size_t size = cls->instance_size + extraBytes + sizeof(intptr_t);
-	intptr_t *addr =
+	intptr_t *addr;
 #ifdef _WIN32
-	// Malloc on Windows doesn't guarantee 32-byte alignment, but we
-	// require this for any class that may contain vectors
-		_aligned_malloc(size, 32);
+	addr = _aligned_malloc(size, OBJC_ALLOC_ALIGN);
+	memset(addr, 0, size);
+#elif OBJC_ALLOC_ALIGN > 16
```
davidchisnall: The 16 here is hard coding an ABI detail. This probably should be `if constexpr (OBJC_ALLOC_ALIGN > alignof(std::max_align_t))`

davidchisnall (approved):

PR #412, opened 2026-08-12
Fix qualifier skipping for nil-message dispatch

HendrikHuebner:
This PR fixes a bug in the message dispatch. We never actually skipped the qualifiers here.

```C
	const char *t = selector->types;
	// Skip type qualifiers
	while ('r' == *t || 'n' == *t || 'N' == *t || 'o' == *t ||
			     'O' == *t || 'R' == *t || 'V' == *t || 'A' == *t) {
		t++;
	}

	switch (selector->types[0]) // t is unused
 i	{ ...
```


davidchisnall (approved):
Oh, bother, I think we fixed this in another place but not here.

davidchisnall:
Ah, no, we fixed it already.

davidchisnall:
(Rebasing turned this into a thing with no changes)

PR #413, opened 2026-08-13
Skip the C++ construct walk with a class flag

DTW-Thalion:
call_cxx_construct_for_class recurses up the superclass chain on every allocation. A class whose superclasses have no .cxx_construct between them walks the chain to do nothing, which is nearly every class in a Foundation workload.

A class flag records that neither a class nor any superclass has one. It is set where the dispatch table is created, after the superclass chain is initialised, and cleared where installMethodInDtable assigns cxx_construct, so a method installed later invalidates it. init_cxx_selectors registers the selector once, in place of a null check on every recursive call.

Minimum of 5 interleaved rounds against master, on whole workloads: Foundation churn under ARC 2.3% faster, allocation of ARC compiled classes about 20% faster, since ARC emits no .cxx_construct and the walk ends immediately. A genomic coder that is AVX2 bound is unchanged.

The matching flag for .cxx_destruct was measured and dropped. It was worth 0.3 to 0.6 points, at or below the spread of the machine, against a branch in every destruction.

ctest passes 198 of 198. Draft because the flag is read on the allocation path without synchronisation, as the other class flags are.

### DIFF
--- a/dtable.c
+++ b/dtable.c
@@ -373,17 +373,12 @@ static BOOL installMethodInDtable(Class class,
 	SparseArrayInsert(dtable, untyped_idx, method);
 #endif
 
-	static SEL cxx_construct, cxx_destruct;
-	if (NULL == cxx_construct)
-	{
-		cxx_construct = sel_registerName(".cxx_construct");
-		cxx_destruct = sel_registerName(".cxx_destruct");
-	}
-	if (selEqualUnTyped(method->selector, cxx_construct))
+	if (selEqualUnTyped(method->selector, cxx_construct_sel))
 	{
 		class->cxx_construct = method->imp;
+		objc_clear_class_flag(class, objc_class_flag_no_cxx_construct);
 	}
-	else if (selEqualUnTyped(method->selector, cxx_destruct))
+	else if (selEqualUnTyped(method->selector, cxx_destruct_sel))
 	{
 		class->cxx_destruct = method->imp;
 	}
@@ -602,6 +597,18 @@ PRIVATE dtable_t create_dtable_for_class(Class class, dtable_t root_dtable)
 			installMethodInDtable(class, dtable, method_at_index(list, i), super_method, YES);
 		}
 		list = list->next;
+	}
+
+	// The superclass chain is fully initialised by this point, so a class
+	// whose superclass has nothing to construct and which has installed no
+	// `.cxx_construct` of its own has nothing to construct either.
+	if ((NULL == super) ||
+	    objc_test_class_flag(super, objc_class_flag_no_cxx_construct))
+	{
+		if (NULL == class->cxx_construct)
+		{
+			objc_set_class_flag(class, objc_class_flag_no_cxx_construct);
+		}
 	}
 
 	return dtable;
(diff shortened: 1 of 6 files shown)


dtable.c
```diff
@@ -604,6 +600,26 @@ PRIVATE dtable_t create_dtable_for_class(Class class, dtable_t root_dtable)
...
+		{
+			objc_set_class_flag(class, objc_class_flag_no_cxx_construct);
+		}
+	}
+	if ((NULL == super) ||
+	    objc_test_class_flag(super, objc_class_flag_no_cxx_destruct))
+	{
+		if (NULL == class->cxx_destruct)
+		{
+			objc_set_class_flag(class, objc_class_flag_no_cxx_destruct);
```
davidchisnall: I'm not quite following the logic here, it looks as if this is never set.  I'd expect each class being resolved to initialise it to the superclass's value and then clear it if the superclass had it set but we found a matching method.

runtime.c
```diff
@@ -46,32 +55,37 @@ PRIVATE void call_cxx_destruct(id obj)
...
 static void call_cxx_construct_for_class(Class cls, id obj)
 {
-	static SEL cxx_construct;
-	if (NULL == cxx_construct)
-	{
-		cxx_construct = sel_registerName(".cxx_construct");
-	}
-
-	if (cls->super_class)
+	if (cls->super_class &&
```
davidchisnall, line 60: Why not do the test in `cls` here and then recurse at the end?

davidchisnall:
This would be good to have some tests for.  These methods are generated by the compiler if an ivar is a C++ class with a constructor or destructor.

Note: I'm not 100% sure that the destruct optimisation will be a net win in modern Objective-C, because the compiler emits the `.cxx_destruct` method for every class with Objective-C object members in ARC, so most classes will have this.  GNUstep Base is unusual in that it's using a 20-year-old Objective-C version.

DTW-Thalion:
> This would be good to have some tests for. These methods are generated by the compiler if an ivar is a C++ class with a constructor or destructor.
> 
> Note: I'm not 100% sure that the destruct optimisation will be a net win in modern Objective-C, because the compiler emits the `.cxx_destruct` method for every class with Objective-C object members in ARC, so most classes will have this. GNUstep Base is unusual in that it's using a 20-year-old Objective-C version.

Agreed - #413 is a work in progress, I worked on this during my flight yesterday, so have some more instrumented tests I need to do - once I am comfortable I'll send you a better report and we can decide.

runtime.c
```diff
@@ -46,32 +50,33 @@ PRIVATE void call_cxx_destruct(id obj)
...
+	    !objc_test_class_flag(cls->super_class, objc_class_flag_no_cxx_construct))
 	{
 		call_cxx_construct_for_class(cls->super_class, obj);
 	}
 	if (cls->cxx_construct)
 	{
-		cls->cxx_construct(obj, cxx_construct);
+		cls->cxx_construct(obj, cxx_construct_sel);
 	}
```
davidchisnall, line 68: I think this can be simplified as:
```suggestion
	if (!objc_test_class_flag(cls, objc_class_flag_no_cxx_construct))
	{
		return;
	}
	cls->cxx_construct(obj, cxx_construct_sel);
	if (cls->super_class != NULL)
	{
		call_cxx_construct_for_class(cls->super_class, obj);
	}
```

PR #417, opened 2026-08-19
Add libc++abi Objective-C exception bridge for WebAssembly

HendrikHuebner:
This PR is the counterpart to https://github.com/llvm/llvm-project/pull/215562.

For native WebAssembly exceptions, Clang emits references to the personality function `__gxx_wasm_personality_v0`.
In order to support Objective-C exception semantics for WASM targets, we could either define our own personality function, or box Objective-C exceptions inside of C++ exceptions and rely on emscriptens libc++abi / exception runtime to handle those correctly for us.

To support Objective-C catch semantics, we need to supply our own type info objects which override the `can_catch` method of libc++abi's internal `__shim_type_info`.

For now, this PR is Apple compatible but does not implement the non-compatible mode.

I'll leave this as future work. I'm not sure whether `AppleCompatibleMode` is currently implemented as it was originally intended, since disabling compatibility would affect both `@catch` and C++ `catch` handler semantics as it is currently written, making them both use static-type matching instead of dynamic-type matching.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,7 +111,11 @@ set(libobjc_CXX_SRCS
 if (WIN32 AND NOT MINGW)
 	list(APPEND libobjc_CXX_SRCS eh_win32_msvc.cc)
 elseif (NOT MINGW)
-	list(APPEND libobjc_C_SRCS eh_personality.c)
+	if (LIBOBJC_WASM)
+		list(APPEND libobjc_CXX_SRCS objc_wasm_eh.cc)
+	else ()
+		list(APPEND libobjc_C_SRCS eh_personality.c)
+	endif ()
 endif ()
 
 find_package(tsl-robin-map)
--- a/objc_wasm_eh.cc
+++ b/objc_wasm_eh.cc
@@ -1,0 +1,141 @@
+#include <cxxabi.h>
+#include <typeinfo>
+#include "objc/runtime.h"
+#include "objc/objc-arc.h"
+#include "visibility.h"
+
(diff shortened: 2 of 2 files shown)


davidchisnall:
This looks very similar to the code in objcxx_eh_mingw, which also wraps C++ exceptions.  Is there a reason to have an entirely new parallel implementation for WAsm?

HendrikHuebner:
Conceptually its doing the same thing as `objcxx_eh_private.h` by wrapping Objective-C exceptions in C++ exceptions.

The problem is that both rely on internal implementation details from different standard libraries.

For targetting emscripten we need our type infos to be subclasses of libc++abi's `__shim_type_info`, which is itself a [subclass](https://github.com/llvm/llvm-project/blob/f652cd614efdc7c10faea86cdc8da7cd96870aa1/libcxxabi/src/private_typeinfo.h#L19) of `std::type_info`.

The existing implementation relies on gcc/libsupc++ implementation details instead, where the `std::type_info` [already contains](https://github.com/gcc-mirror/gcc/blob/74b3180bb616fbc35d1b655da75d8f6d8b39d368/libstdc%2B%2B-v3/libsupc%2B%2B/typeinfo#L93) the `__do_catch` member and whatever else we need to override.

I don't think these implementations can be merged, sadly.

Edit: maybe some things could be reused, such as the implementation of `__do_catch`/`can_catch`?

HendrikHuebner:
The Clang side of this has landed: https://github.com/llvm/llvm-project/pull/215562

PR #418, opened 2026-08-19
Cast IMPs to the correct signature before calling

HendrikHuebner:
This PR ensures we do not call any raw IMP function pointers without casting them to the correct type first.

Calling a funtion pointer with a mismatching signature is technically UB I believe but tends to work just fine on many native platforms.
With WebAssembly, this is a problem however, because WASM indirect calls require an exact signature match. This can cause runtime failures.


arc.mm
```diff
@@ -250,9 +250,12 @@ static TLS_CALLBACK(cleanupPools)(struct arc_tls* tls)
...
 static Class AutoreleasePool;
-static IMP NewAutoreleasePool;
-static IMP DeleteAutoreleasePool;
-static IMP AutoreleaseAdd;
+typedef id (*NewAutoreleasePoolIMP)(id, SEL);
+typedef void (*DeleteAutoreleasePoolIMP)(id, SEL);
+typedef void (*AutoreleaseAddIMP)(id, SEL, id);
+static NewAutoreleasePoolIMP NewAutoreleasePool;
+static DeleteAutoreleasePoolIMP DeleteAutoreleasePool;
+static AutoreleaseAddIMP AutoreleaseAdd;
```
davidchisnall: How about something like this:

```suggestion
template<typename Return, typename... Arguments>
using Selector = Return(*)(id, SEL, Arguments...);

static Selector<id> NewAutoreleasePool;
static Selector<void> DeleteAutoreleasePoolIMP DeleteAutoreleasePool;
static Selector<void, id> AutoreleaseAdd;
```

arc.mm
```diff
@@ -566,12 +569,15 @@ static inline void initAutorelease(void)
...
 				[AutoreleasePool class];
-				NewAutoreleasePool = class_getMethodImplementation(object_getClass(AutoreleasePool),
-				                                                   SELECTOR(new));
-				DeleteAutoreleasePool = class_getMethodImplementation(AutoreleasePool,
-				                                                      SELECTOR(release));
-				AutoreleaseAdd = class_getMethodImplementation(object_getClass(AutoreleasePool),
-				                                               SELECTOR(addObject:));
+				NewAutoreleasePool = reinterpret_cast<NewAutoreleasePoolIMP>(
+					class_getMethodImplementation(object_getClass(AutoreleasePool),
+					                             SELECTOR(new)));
```
davidchisnall: Can we consolidate the casts with something like:

```c++
auto storeSelector = [](auto &target, IMP imp)
    {
         target = reinterpret_cast<std::remove_reference_t<decltype(target)>>(imp);
    };
```

runtime.c
```diff
@@ -46,7 +49,8 @@ PRIVATE void call_cxx_destruct(id obj)
 		cls = cls->super_class;
 		if (currentClass->cxx_destruct)
```
davidchisnall, line 47: We should probably change the type of these from IMP and do the cast when we set them.

HendrikHuebner:
updated

davidchisnall (approved):

davidchisnall:
Please can you squash for merging?

HendrikHuebner:
Done. I don't think I have permissions to merge PRs myself.

